PR #5, opened 2018-04-06
Add more protos to the repo following the spec

jvican:
This is still a work in progress. So far, this PR:

* Add endpoints for new bsp endpoints
* Add ScalaPlatform to build target
* Add scala test classes protos
* Add clarification to the docs that compile is intransitive


docs/bsp.md
```diff
@@ -435,6 +435,8 @@ trait DependencySourcesItem {
 ### 1.6.7. Compile Request
 
 The compile build target request is sent from the client to the server to compile the given list of build targets.
+The compilation request will not compile these projects transitively.
```
olafurpg: Can you elaborate on this?

jvican: Imagine the following scenario:

laughedelic: 💭

olafurpg:
Ping, any update on this?

src/main/protobuf/bsp.proto
```diff
@@ -135,17 +150,43 @@ message ScalaBuildTarget {
     string scalaOrganization = 1;
     string scalaVersion = 2;
     string scalaBinaryVersion = 3;
+    ScalaPlatform platform = 4;
```
olafurpg: Seems off-sync with the spec. It's tedious to keep the two synchronized but I think it's worth the effort.

docs/bsp.md
```diff
@@ -253,25 +248,37 @@ trait InitializeBuildResult {
 }
 
 trait BuildServerCapabilities {
-  /** The server can compile targets via method buildTarget/compile */
-  def compileProvider: Boolean
-  /** The server can test targets via method buildTarget/test */
-  def testProvider: Boolean
+  /** The languages the server supports compilation via method buildTarget/compile. */
+  def compileProvider: {
+    val languageIds: List[String]
```
olafurpg: s/val/def/

docs/bsp.md
```diff
@@ -341,31 +348,31 @@ Notification:
 
 ```scala
 trait ShowMessageParams {
-	/** The message type. See {@link MessageType}. */
-	def type: Int
+  /** The message type. See {@link MessageType}. */
+  def `type`: Int
```
olafurpg: Let's keep `type` un-backticked to avoid confusion for non-Scala readers that the backticks are not expected in JSON object keys.

docs/bsp.md
```diff
@@ -652,32 +658,38 @@ trait TestParams {
...
+  
   /** The total number of successful tests. */
   def passed: Int
 
   /** The total number of failed tests. */
   def failed: Int
 
-  /** The total number of milliseconds it took to run the tests.
-    * Should not include compile times. */
+  /** The total number of milliseconds it took to only run the tests. */
```
olafurpg: I think the previous docstring was clearer, it's good to explicitly say this should not include compile or packaging times.

docs/bsp.md
```diff
@@ -377,16 +380,21 @@ object MessageType {
 }
 ```
 
+A `build/showMessage` notification is similar to LSP's `window/showMessage`, except for a few new
+additions.
+
+The `id` and optional `parentId` fields allow clients to structure logs in a hierarchical way (in a
+tree fashion, with dropdowns, et cetera) to ease readability.
```
olafurpg: s/, et cetera//

Same goes for "obviously" and "clearly", makes readers feel dumb when they don't understand

docs/bsp.md
```diff
@@ -352,10 +352,13 @@ trait ShowMessageParams {
   def `type`: Int
 
   /** The message id. */
-  def id: Long
+  def id: string
```
olafurpg: Should this be called `taskId` instead? And `parentTaskId`, respectively. The docstrings should have the explanation below on what the difference between "task id" and "request id" is.

olafurpg: It would also be good to mention that this is is generated by the server, not the client.

olafurpg: Should this be optional?

docs/bsp.md
```diff
@@ -684,7 +725,12 @@ The server communicates during the initialize handshake whether this method is s
 
 ```scala
 trait TestParams {
+  /** A sequence of build targets to test. */
   def targets: List[BuildTargetIdentifier]
+  
+  /** An optional number uniquely identifying a client request. */
```
olafurpg: > A unique identifier generated by the client to identify this request.
> The server may include this id as `requestId` in notifications such as `build/compileReport`

docs/bsp.md
```diff
@@ -723,9 +776,11 @@ trait TestReport {
 The `TestReport` notification will be sent as the test execution of build targets completes.
 
 This request may trigger a compilation on the selected build targets. The server is free to send any
-number of `buildTarget/publishDiagnostics` and `build/logMessage` notifications during compilation
-before completing the response. The client is free to forward these messages to the LSP editor
-client.
+number of `build/publishDiagnostics` and `build/logMessage` notifications during compilation before
```
olafurpg: Include `build/compileReport` as another example

docs/bsp.md
```diff
@@ -803,42 +858,57 @@ trait ScalacOptionsItem {
 
 #### 1.7.1.3. Scala Test Classes Request
 
-The build target scalac options request is sent from the client to the server to query for the list of fully qualified names of test clases in a given list of targets.
-This method can for example be used by a language server to attach a "Run test suite" button above the definition of a test suite via `textDocument/codeLens`.
-To render the code lens, the language server needs to map the fully qualified names of the test targets to the defining source file via `textDocument/definition`.
-Then, once users click on the button, the language server can pass the fully qualified name of the test class as an argument to the `buildTarget/test` request.
+The build target scalac options request is sent from the client to the server to query for the list
```
olafurpg: s/The build target scalac options request/The Scala test classes request/

docs/bsp.md
```diff
@@ -803,42 +858,57 @@ trait ScalacOptionsItem {
...
+of fully qualified names of test clases in a given list of targets. This method can for example be
```
olafurpg: End the paragraph at "list of targets" and start a new paragraph at "This method can for example"

jvican:
Just addressed all feedback, feel free to refine the introduction and motivation. We're still far from having something as convincing as I'd like them to be.
  
Did you have a look at the addition `buildTarget/run` and `buildTarget/scalaMainClasses`?

jvican:
@jastice Feel free to have a look at the changes, your input is appreciated :smile: I haven't yet implemented in the bloop side but I think that the sbt language extension has all we need to have better build.sbt file support in IntelliJ, could you confirm?

jastice:
@jvican From the description it's not clear to me: is `SbtBuildTarget` supplied for any target used by sbt? Is `parent` supposed to be the meta build of that target or the other way around?

Minimum information missing: sbt version (required for correct highlighting)
probably helpful: sbt project uri/id, configuration, project base dir. these could be used to merge targets back to intellij modules where necessary.

jvican:
@jastice Thanks for having a look!
  
The child (a normal build) has a reference to the parent (meta-build), and so on. The idea is that there will be an sbt build target per sbt project (including meta projects) so that IntelliJ can import them correctly.
  
You're right that the sbt version is missing. The project URI is already in the `BuildTarget`. The way the structure works is that the `BuildTarget` has a data field that contains the language specific information, in this case `SbtBuildTarget`.
  
What I don't understand is why you need configurations. Isn't the configuration always going to be `Compile`?

jastice:
Is the BuildTarget uri the exact project uri? how is the sbt project id included?

In bloop you have separate `name` and `name-test` targets that I combine back into a single IDEA module with compile and test sources/dependencies/etc respectively. How do you intend to do this for sbt?

jvican:
> Is the BuildTarget uri the exact project uri? how is the sbt project id included?
  
The build target URI will be the one of the sbt project URI, like `/data/rw/code/scala/foo/project`. What do you mean by sbt project id?
  
> In bloop you have separate name and name-test targets that I combine back into a single IDEA module with compile and test sources/dependencies/etc respectively. How do you intend to do this for sbt?
  
As there's only one configuration per sbt project, I was thinking of just having one IDEA module for the sbt build. Is that possible?

jvican:
This is ready for review again @olafurpg

docs/bsp.md
```diff
@@ -864,6 +864,9 @@ trait ScalaBuildTarget {
 
   /** The target platform for this target */
   def platform: Int
+  
+  /** A sequence of Scala jars. */
```
olafurpg: Can you elaborate on what jars these are? scala-library, scala-reflect, scala-compiler?

olafurpg: the `scalac` command line interface does not have a --scala-jars flag so it's not obvious to me  what this is

docs/bsp.md
```diff
@@ -370,20 +367,36 @@ where `MessageType` is defined as follows:
...
+
   /** A log message. */
   final val Log = 4
 }
 ```
 
+and a `TreeStructure` is defined as:
+
+```scala
+trait TreeStructure {
```
olafurpg: I'm not so keen on this name. Given that `TreeStructure` is only used in two messages, can we revert this commit and keep `nodeId` and `parentId` inlined?

olafurpg: If we keep it, this should be included in the "base data structures" section.

docs/bsp.md
```diff
@@ -147,6 +147,10 @@ trait BuildTarget {
     * The id.uri is used if None. */
   def displayName: Option[String]
 
+  /** The type of build target. Useful for an IDE to show
+    * targets with different kinds in the UI. */
+  def kind: Int
```
olafurpg: `App` is missing. What about projects that are for example of kind bench AND test?  I think it would be better to have `labels: List[String]` and leave it to clients and servers to come up with meaningful names.

olafurpg:
Whoop, accidentally closed the PR, sorry.

I am ok with keeping `BuildTarget.kind` if we make it a `List[String]`, I'm convinced it's useful information.

docs/bsp.md
```diff
@@ -159,6 +163,13 @@ trait BuildTarget {
   def data: Option[Json] // Note, matches `any` in the LSP.
 }
 
+object BuildTargetKind {
+  final val Library = 1
+  final val Test = 2
+  final val IntegrationTest = 3
+  final val Bench = 4
```
olafurpg: If we keep kind as a single element, 

- it should be a string instead of int, similarly as language IDs are strings
- we are missing `App` and in the future more categories may come up like `iOS`, `Android`

I personally lean towards turning `kind: Int` into `labels: List[String]` where we provide a builtin set of strings like we do with languages.

olafurpg: After more thinking and offline discussions, I'm in favor of keeping `BuildTarget.kind: Int` but we should add `final val App = ...`

jvican:
Feedback addressed, thanks for your feedback @olafurpg :pray: I'm merging.

PR #11, opened 2018-05-28
Fix link to protocol buffers definitions

propensive:


olafurpg:
BTW, can you please create branches in your fork? I'd like to try and have only `master` when people clone origin.

PR #13, opened 2018-06-10
Bring spec and implementation up-to-date

jvican:
This PR removes the scalapb dependency (to reduce the amount of dependencies in
our classpath) and implements the previous protobuf file with circe.

It also brings both the implementation and the spec up to date after the heavy
changes in https://github.com/scalacenter/bsp/pull/5, and refines the rough
edges of the specification.

This is the list of todo items to do before merging this.

- [ ] Finish the circe encoder/decoders for sealed traits.
- [ ] Change `List[String]` by `List[Uri]`


jvican:
This is ready for review, @olafurpg do you mind taking a look?

docs/bsp.md
```diff
@@ -279,26 +301,26 @@ trait BuildServerCapabilities {
   
   /** The server can provide a list of targets that contain a
     * single text document via the method textDocument/buildTargets */
-  def textDocumentBuildTargetsProvider: Boolean
+  def providesTextDocumentBuildTargets: Boolean
```
olafurpg: I'm concerned of diverging from LSP here on naming conventions. I would prefer to keep the name and turn `Boolean` into an object, like LSP does.

If we keep this change, we should synchronize the name with `canTest`/`canRun`/`canCompile`.

docs/bsp.md
```diff
@@ -725,7 +735,7 @@ trait CompileReport {
   def warnings: Int
 
   /** The total number of milliseconds it took to compile the target. */
-  def time: Option[Int]
+  def time: Option[Long]
```
olafurpg: Long does fit in a JSON number. Let's turn this into a Double instead.

docs/bsp.md
```diff
@@ -792,9 +802,21 @@ trait TestReport {
...
+
+  /** The total number of skipped tests. */
+  def skipped: Int
+
+  /** The total number of pending tests. */
+  def pending: Int
 
   /** The total number of milliseconds tests take to run (e.g. doesn't include compile times). */
-  def time: Option[Int]
+  def time: Option[Long]
```
olafurpg: s/Long/Double/

docs/bsp.md
```diff
@@ -841,7 +863,20 @@ Response:
...
+  def statusCode: Int
+}
+
+object StatusCode {
+  /** Execution was successful. */
+  val Ok = 0
+  /** Execution failed. */
+  val Error = 1
+  /** Execution was cancelled. */
+  val Cancelled = -1
```
olafurpg: Is there a reason this is not `= 2`?

docs/bsp.md
```diff
@@ -1062,23 +1097,22 @@ trait SbtBuildTarget {
...
   
-  /** The scala version to compile this target. */
-  def scalaVersion: String
-  
-  /** The sbt version. Useful to support version-dependent syntax. */
-  def sbtVersion: String
-
-  /** A sequence of Scala jars. */
-  def scalaJars: List[String]
+  /** The Scala build target for this sbt build target. */
```
olafurpg: Can you elaborate on this docstring? It's not clear to me what this is for.

src/main/scala/ch/epfl/scala/bsp/Bsp.scala
```diff
@@ -0,0 +1,401 @@
+package ch.epfl.scala.bsp
+
+import java.net.{URI, URISyntaxException}
+
+import io.circe.Decoder.Result
+import io.circe.{Decoder, DecodingFailure, HCursor, Json, RootEncoder}
+import io.circe.generic.JsonCodec
+
+final case class Uri private[Uri](val s: String)
```
olafurpg: s/[Uri]// ?

I think the constructor is still accessible.

Also, what does `s` stand for in `val s: String`? I would expect `value: String`

src/main/scala/ch/epfl/scala/bsp/Bsp.scala
```diff
@@ -0,0 +1,401 @@
...
+  case object Log extends MessageType(4)
+
+  implicit val messageTypeEncoder: RootEncoder[MessageType] = new RootEncoder[MessageType] {
+    override def apply(a: MessageType): Json = Json.fromInt(a.id)
+  }
+
+  implicit val messageTypeDecoder: Decoder[MessageType] = new Decoder[MessageType] {
+    override def apply(c: HCursor): Result[MessageType] = {
+      c.as[Int].flatMap {
+        case 1 => Right(Error)
```
olafurpg: I think we should abstract over this boilerplate with a simple function

```scala
def enumDecoder[T](allValues: List[T], getId: T => Int, onError: Int => String): Decoder[T] = ...
def enumEncoder[T](allValues: List[T], getId: T => Int): Encoder[T] = ...

sealed trait MessageType
object MessageType {
  val all = List(Error, Warning, Info)
  implicit val decoder  = enumDecoder(all, _.id, n => s"Bad id $n")
```

That avoids the risk of doing a copy-paste mistake in the `case MAGIC_CONSTANT` pattern match.

Alternatively, we use enumeratum library in lsp4s to avoid doing this by hand

https://github.com/scalameta/lsp4s/blob/de927af3a9e9664a650b8247c9280e777388bda1/lsp4s/src/main/scala/scala/meta/lsp/Enums.scala#L5-L17

It's a relatively small dependency (48K).

src/main/scala/ch/epfl/scala/bsp/package.scala
```diff
@@ -0,0 +1,5 @@
+package ch.epfl.scala
+
+package object bsp {
```
olafurpg: Is this package object necessary?

docs/bsp.md
```diff
@@ -1062,23 +1097,22 @@ trait SbtBuildTarget {
...
+  def scalaBuildTarget: ScalaBuildTarget
 }
 ```
 
+where `parent` points to the sbt metabuild of this target (if any).
+
 #### 1.7.2.1. Sbt Build Target
 
 ## 1.8. Appendix
```
olafurpg: Looks like the Appendix is outdated now.

docs/bsp.md
jvican: I changed it because I didn't know LSP used `provider` when the type of the field was `boolean`, I agree with you that we should follow the convention there so I'll revert this change.

docs/bsp.md
jvican: I'm not sure we should make this a double since it will store most likely ms. What I want here is a type representing the biggest integer possible.
  
The question is: should we type things as in Scala or as JSON allows? Another way to phrase it: is `Int` everywhere in the proposal representing an integer following the JSON spec?
  
If that's the case, we should probably add a paragraph explaining that.

src/main/scala/ch/epfl/scala/bsp/Bsp.scala
jvican: I agree this is much nicer, but let's do this after 1.0 is out. It won't change semantics and it only affects the implementation and maintenance of this tiny library.

src/main/scala/ch/epfl/scala/bsp/package.scala
jvican: No it isn't!

docs/bsp.md
```diff
@@ -192,6 +193,25 @@ trait BuildTargetIdentifer {
...
+
+The Hierarchical Id allows clients to uniquely identify a resource and establish a client-parent
+relationship with another id.
+
+```scala
+trait HierarchicalId {
+  /** The id */
+  def id: String
+
+  /** The parent id. */
```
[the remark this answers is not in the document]
jvican: This should be `Option[String]` indeed

docs/bsp.md
```diff
@@ -192,6 +193,25 @@ trait BuildTargetIdentifer {
 }
 ```
 
+### 1.5.3. Hierarchical Id
```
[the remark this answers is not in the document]
jvican: I'm fine with this name to avoid `.id.id`

src/main/scala/ch/epfl/scala/bsp/Bsp.scala
jvican: The constructor is only accessible inside the class and companion `Uri`. I'm changing the name `s` to `value`.

docs/bsp.md
```diff
@@ -1106,34 +1106,25 @@ trait SbtBuildTarget {
...
-### 1.8.2. Scala Bindings
+### 1.8.1. Scala Bindings
 
 A Scala library implementation of this communication protocol is available in this repository.
-The public API of this library currently has three direct Scala dependencies:
+The public API of this library currently has two direct Scala dependencies:
 
-* ScalaPB - for generation of Scala sources from protobuf schema
 * Monix - for asynchronous programming primitives
 * Circe - for JSON serialization and parsing of protocol data structures
```
olafurpg: Missing scribe

jvican:
Done @olafurpg, thank you :smile:

docs/bsp.md
```diff
@@ -212,6 +212,24 @@ An example of use of hierarchical ids is logs, where BSP clients can use the hie
 to improve their readability in the user interface. Clients can show logs in a tree fashion, for
 example, or with dropdowns.
 
+### 1.5.4. DocumentUri
+
+A document URI uniquely identifies a document.
+
+```scala
+/** A id that is a valid URI. https://tools.ietf.org/html/rfc3986 */
```
olafurpg: Should we get rid of `DocumentUri` in favor of `Uri`? What is their difference? The spec does not say.

jvican:
Alright, these are all the changes I envision for the 1.0 release, we should leave the remaining tickets in our issue tracker for 2.0.

docs/bsp.md
```diff
@@ -187,11 +188,48 @@ A unique identifier for a target.
...
+type DocumentUri = String
+```
+
+### 1.5.5. Uri
+
+A URI uniquely identifies a resource.
+
+```scala
+/** A id that is a valid URI. https://tools.ietf.org/html/rfc3986 */
```
olafurpg: > /** A resource identifier that is a valid URI according to rfc3986: https:/... */

docs/bsp.md
```diff
@@ -187,11 +188,48 @@ A unique identifier for a target.
...
+A document URI uniquely identifies a document.
+
+```scala
+/** A id that is a valid URI. https://tools.ietf.org/html/rfc3986 */
+type DocumentUri = String
+```
+
+### 1.5.5. Uri
+
+A URI uniquely identifies a resource.
```
olafurpg: Remove this in favor of the docstring below

docs/bsp.md
```diff
@@ -279,26 +322,52 @@ trait BuildServerCapabilities {
...
+    * to notify events about via `workspace/didChangeWatchedFiles`. */
+  def watchers: List[FileSystemWatcher]
+}
+```
+
+where `FileSystemWatcher` is described as follows:
+
+```scala
+trait FileSystemWatcher {
+   /** The glob pattern to watch in all the workspace. */
```
olafurpg: Can you document the expected syntax? Are separators cross platform?

docs/bsp.md
```diff
@@ -279,26 +322,52 @@ trait BuildServerCapabilities {
...
+  def providesResources: Boolean
   
   /** The server sends notifications to the client on build
     * target change events via buildTarget/didChange */
-  def buildTargetChangedProvider: Boolean
+  def providesBuildTargetChanged: Boolean
+  
+  /** The file system watchers that the server wants the client
+    * to notify events about via `workspace/didChangeWatchedFiles`. */
+  def watchers: List[FileSystemWatcher]
```
olafurpg: Is this a server capability? How does one dynamically update the patterns?

I would try to copy what LSP does,

- https://microsoft.github.io/language-server-protocol/specification#didchangewatchedfiles-notification-arrow_right
- https://microsoft.github.io/language-server-protocol/specification#register-capability-arrow_right_hook

This is one reason why I'm concerned about the `s/dependencySourcesProvider/providesDependencySources/` renaming since it moves from how LSP works.

docs/bsp.md
```diff
@@ -192,6 +193,25 @@ trait BuildTargetIdentifer {
...
+The Hierarchical Id allows clients to uniquely identify a resource and establish a client-parent
+relationship with another id.
+
+```scala
+trait HierarchicalId {
+  /** The id */
+  def id: String
+
+  /** The parent id. */
+  def parentId: String
```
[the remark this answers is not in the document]
olafurpg: This is still `parentId`

docs/bsp.md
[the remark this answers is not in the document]
olafurpg: Are you in favor of `task: Option[TaskId]` or do want to keep `HierarchicalId`?

docs/bsp.md
jvican: > How does one dynamically update the patterns?

docs/bsp.md
jvican: It's the same, wanted to keep it to be more similar to LSP, but I agree it's redundant.

jvican:
Alright, I updated the TOC and fixed the remaining issues. Should be ready for another pass.

docs/bsp.md
olafurpg: > Currently, every change in a file watcher requires a new connection.

That seems like a strict limitation, I can imagine that it's normal to adjust the file watching patterns based on the build target sources. I suggest we come up with a design that supports dynamic updates to the file watching patterns or we remove it from the spec. I don't think it's fine to introduce this feature when we are sure it will break in the near future.

olafurpg: btw, we don't need to implement dynamic registration for file watching. I personally don't really understand how dynamic registration works in LSP and it's not supported by clients like atom (my hypothesis is that dynamic capabilities are considered too complex for the simple problem).

I think it would make sense to have a request to register interest in file watching a pattern and the response includes an ID. There would then be another request to cancel the file watcher.

```
build/registerFileWatcher
Request: { globs: [ "**/*.scala" ] }
Response: { id: ID }

... (sequence of didChangeWatchedFiles notifications)

build/cancelFileWatcher
Request: { id: ID }
Response: { cancelled: true/false }
```

docs/bsp.md
```diff
@@ -792,6 +884,18 @@ trait TestReport {
...
   /** The total number of failed tests. */
   def failed: Int
+  
+  /** The total number of ignored tests. */
+  def ignored: Int
+
+  /** The total number of cancelled tests. */
+  def cancelled: Int
+
+  /** The total number of skipped tests. */
```
smarter: What's the difference between ignored and skipped?

docs/bsp.md
```diff
@@ -792,6 +884,18 @@ trait TestReport {
...
+  def skipped: Int
+
+  /** The total number of pending tests. */
```
smarter: What does pending mean for a test?

docs/bsp.md
```diff
@@ -841,7 +945,20 @@ Response:
 trait RunResult {
   /** An optional request id to know the origin of this report. */
   def requestId: Option[String] 
+  
+  /** An status code for the execution. Allows to use POSIX run c */
```
smarter: "run c"?

docs/bsp.md
```diff
@@ -1035,8 +1152,8 @@ trait ScalaMainClass {
   /** The user arguments to the main entrypoint. */
   def arguments: List[String]
   
-  /** The java options the application. */
-  def javaOptions: List[String]
+  /** The jvm options the application. */
```
smarter: _for_ the application

docs/bsp.md
jvican: Very nice proposal, I like it :+1: I just pushed a commit implementing it.

docs/bsp.md
jvican: I think ignored are tests ignored by the user (with `@Ignored`), whereas skipped are tests not run by the build tool. I got these fields from sbt's test interface design.

docs/bsp.md
jvican: I agree it doesn't make much sense, I remove it

jvican: I agree it doesn't make much sense, I remove it

docs/bsp.md
```diff
@@ -502,7 +525,118 @@ trait WorkspaceBuildTargetsResult {
...
+
+Request:
+
+* method: `build/registerFileWatcher`
+* params: `RegisterFileWatcherParams`, defined as follows
+
+```scala
+trait RegisterFileWatcherParams {
+   /** The glob pattern to watch in all the workspace.
+     * Syntax is implementation specific. */
```
olafurpg: Let's try to make an effort to define the syntax of this magic string 

> The glob pattern to watch in the workspace, for example `src/**/*.{scala,js}`.
> - `*` Matches 0 or more characters in a single path portion
> - `**` matches zero or more path portions
> - `*.{scala,js}` expands to two patterns: `*.scala` and `*.js`
> - Forward slashes `/` are always treated as path separators, even on Windows.
>
> Further details are implementation defined.

jvican: Sounds reasonable to me :+1: Pity that LSP doesn't define on its own.

docs/bsp.md
```diff
@@ -181,33 +182,62 @@ trait BuildTargetCapabilities {
...
+The task Id allows clients to uniquely identify a resource and establish a client-parent
+relationship with another id.
+
+```scala
+trait TaskId {
+  /** The id */
+  def id: String
+
+  /** The parent id. */
+  def parent: String
```
olafurpg: Is this supposed to be an `Option[String]`?

docs/bsp.md
```diff
@@ -502,7 +525,118 @@ trait WorkspaceBuildTargetsResult {
...
+  def changes: List[FileEvent]
+}
+```
+
+Where `FileEvent`s are described as follows:
+
+```scala
+trait FileEvent {
+  /** The file's URI. */
+  def uri: uri
```
olafurpg: s/uri/Uri/

olafurpg:
Also great work refactoring out the protobuf generation, that simplifies a lot :smile:

PR #16, opened 2018-06-12
Fix some oversights in the implementation

jvican:
These changes will be shipped in M2.


jvican:
Thanks for the swift review :pray:

PR #17, opened 2018-06-13
Add `definedTargets` to the sbt build target

jvican:
This information is critical to tie together sbt build targets to a
series of scala projects and to reconstruct the sbt meta build graph.

It also fixes a typo in the name of the function for
`dependencyResources` and clarifies the spec of build target identifier.

### DIFF
--- a/docs/bsp.md
+++ b/docs/bsp.md
@@ -1247,7 +1247,7 @@ trait SbtBuildTarget {
   /** The build targets defined by the sbt build represented
     * in this target. It can contain normal project targets or sbt
     * build targets if this target represents an sbt meta-meta build. */
-  def definedTargets: List[BuildTargetIdentifier]
+  def children: List[BuildTargetIdentifier]
 }
 ```
 
@@ -1258,7 +1258,7 @@ The sbt build target for `M1` will have `A` and `B` as the defined targets and `
 Similarly, the sbt build target for `M2` will have `M1` as the defined target and no parent.
 
 Clients can use this information to reconstruct the tree of sbt meta builds. The
-`parent` information can be defined from `definedTargets` but it's provided by the server to
+`parent` information can be defined from `children` but it's provided by the server to
 simplify the data processing on the client side.
 
 ## Appendix
--- a/docs/bsp.md
+++ b/docs/bsp.md
@@ -184,7 +184,7 @@ trait BuildTargetCapabilities {
 
 ### Build Target Identifier
 
-A unique identifier for a target.
+A unique identifier for a target, usually the project base directory.
 
 ```scala
 trait BuildTargetIdentifer {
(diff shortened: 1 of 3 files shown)


docs/bsp.md
```diff
@@ -1240,13 +1237,23 @@ trait SbtBuildTarget {
...
+  
+  /** The build targets defined by the sbt build represented
+    * in this target. It can contain scala targets or sbt build
+    * targets if this target represents an sbt meta-meta build. */
+  def definedTargets: List[Json]
 }
 ```
 
-where `parent` points to the sbt metabuild of this target (if any).
+Clients can use this information to reconstruct the tree of sbt meta builds in a simple way. The
```
olafurpg: Remove : in a simple way.

docs/bsp.md
```diff
@@ -1240,13 +1237,23 @@ trait SbtBuildTarget {
...
+   * version and scala jars used by this sbt version. */
   def scalaBuildTarget: ScalaBuildTarget
+  
+  /** An optional parent if the target has an sbt meta project. */
+  def parent: Option[BuildTargetIdentifier]
+  
+  /** The build targets defined by the sbt build represented
+    * in this target. It can contain scala targets or sbt build
+    * targets if this target represents an sbt meta-meta build. */
+  def definedTargets: List[Json]
```
olafurpg: Can we type this better? For example separate into two fields : definedSbtTargets and definedScalaTargets

docs/bsp.md
```diff
@@ -1240,13 +1237,23 @@ trait SbtBuildTarget {
...
-  /** The Scala build target associated with this sbt build.
-    * It contains the scala version and the scala jars used. */
+  /** The Scala build target describing the scala
+   * version and scala jars used by this sbt version. */
   def scalaBuildTarget: ScalaBuildTarget
+  
+  /** An optional parent if the target has an sbt meta project. */
+  def parent: Option[BuildTargetIdentifier]
+  
+  /** The build targets defined by the sbt build represented
```
olafurpg: It's not clear to me what this means. Can you give a basic example or reword?

docs/bsp.md
jvican: This should have been `BuildTargetIdentifier` from the start. Now it is, it's more typed and represents better that it can create scala and sbt build targets. I prefer to keep this rather than the two independent fields.

docs/bsp.md
```diff
@@ -1240,13 +1237,29 @@ trait SbtBuildTarget {
...
+    * in this target. It can contain normal project targets or sbt
+    * build targets if this target represents an sbt meta-meta build. */
+  def definedTargets: List[BuildTargetIdentifier]
```
olafurpg: Should this be named `children`? Judging by the description it seems like the inverse of `parent`

jvican: I agree with you this is more consistent with `parent`, I've just renamed it

PR #18, opened 2018-06-14
Replace `requestId` by `originId`

jvican:
Let's make this change so that it's clearer that this id does not intend
to replace the Jsonrpc id. The name `requestId` was misleading.


docs/bsp.md
```diff
@@ -829,7 +829,7 @@ trait CompileParams {
   
   /** A unique identifier generated by the client to identify this request.
     * The server may include this id in triggered notifications or responses. */
```
olafurpg: s/or responses//

JSON-RPC responses already include an ID matching the request. The `originID` is only relevant for notifications.

docs/bsp.md
```diff
@@ -844,7 +844,7 @@ Response:
 ```scala
 trait CompileResult {
   /** An optional request id to know the origin of this report. */
-  def requestId: Option[String]
+  def originId: Option[String]
```
olafurpg: We should not include the `originId` in the response message. The JSON-RPC response will already match the request

jvican: Yes, it's mostly useful to notifications because JSON-RPC doesn't have ids for notifications that link them to the original ID. I'm not sure we should remove it though, it seems to me it may still be useful for a client and it would remove a little bit the consistency of the API.
  
Say a client has indexed some information based on the origin id in a map, and you want to access it when the response is completed so that it can act on it. Without `originId` in the response, the user cannot do so because it doesn't have access to the JSON-RPC request id.

olafurpg: The client already has access to the `originId` in the response because the response is handled after submitting the request

```scala
val originId = ???
val request = CompileParams(originId)
globalMap(originId) = metadata(...)
client.compile(request).onComplete { response =>
  globalMap(originId)...
}
```

I don't feel strongly about this so we can keep it unchanged. It doesn't hurt to include the `originId` in the reponse but I feel like it's a bit redundant.

PR #25, opened 2018-06-20
Switch from circe-generic to circe-derivation

andyscott:
Uses circe-derivation instead of circe-generic. See https://github.com/scalameta/lsp4s/pull/9 for the same change in the LSP project and some additional context.

Note: Currently circe-generic will be pulled in transitively until we update the lsp4s dependency version to a release cut with the linked PR.


jvican:
Thanks Andy :smile:

andyscott:
Sure thing!

Also: rebased on the changes in #26. Hopefully the build works.

gabro:
Heads up: jsonrpc 0.2.0 is out :)

olafurpg:
Release notes https://github.com/scalameta/lsp4s/releases/tag/v0.2.0

andyscott:
I can open a separate PR for the LSP version bump. I like to keep changes as logical as I can, when feasible.

gabro:
Yup, my comment was about the circe-generic dependency which is now gone even transitively in 0.2.0.

andyscott:
Aah got it. I went ahead and opened the other PR, which should get a rebase before merging. I can take care of that after this is merged.

PR #33, opened 2018-10-16
Implement bsp4j module based on org.eclipse.lsp4j.jsonrpc

olafurpg:
Add a new module `ch.epfl.scala:bsp4j` that is written in pure Java in
the same style as the lsp4j module. The benefit of this module compared
to `ch.epfl.scala:bsp_2.12` is that it has a lower dependency footprint.

Includes a full end-to-end integration test with bloop that stresses the
current implemented endpoints in bloop.  The module implements the full
spec but we can't test all endpoints yet since no server implements
them.

Future work:
- add docstrings for the data structures and endpoints
- add `@NonNull` where necessary

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java
@@ -8,16 +8,16 @@ import java.util.concurrent.CompletableFuture;
 public interface BuildClient {
 
     @JsonNotification("build/showMessage")
-    void showMessage(ShowMessageParams params);
+    void onShowMessage(ShowMessageParams params);
 
     @JsonNotification("build/logMessage")
-    void logMessage(LogMessageParams params);
+    void onLogMessage(LogMessageParams params);
 
     @JsonNotification("build/publishDiagnostics")
-    void publishDiagnostics(PublishDiagnosticsParams params);
+    void onPublishDiagnostics(PublishDiagnosticsParams params);
 
     @JsonNotification("buildTarget/didChange")
-    void didChangeBuildTarget(DidChangeBuildTarget params);
+    void onBuildTargetChanged(DidChangeBuildTarget params);
 
     @JsonRequest("build/registerFileWatcher")
     CompletableFuture<RegisterFileWatcherResult> registerFileWatcher(RegisterFileWatcherParams params);
@@ -26,9 +26,9 @@ public interface BuildClient {
     CompletableFuture<CancelFileWatcherResult> cancelFileWatcher(CancelFileWatcherParams params);
 
     @JsonNotification("buildTarget/compileReport")
-    void compileReport(CompileReport params);
+    void onCompileReport(CompileReport params);
 
-    default void connect(BuildServer server) {
+    default void onConnectWithServer(BuildServer server) {
 
     }
 
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java
@@ -8,25 +8,25 @@ import java.util.concurrent.CompletableFuture;
 public interface BuildClient {
 
     @JsonNotification("build/showMessage")
-    void onShowMessage(ShowMessageParams params);
+    void onBuildShowMessage(ShowMessageParams params);
 
     @JsonNotification("build/logMessage")
-    void onLogMessage(LogMessageParams params);
+    void onBuildLogMessage(LogMessageParams params);
 
     @JsonNotification("build/publishDiagnostics")
-    void onPublishDiagnostics(PublishDiagnosticsParams params);
+    void onBuildPublishDiagnostics(PublishDiagnosticsParams params);
 
     @JsonNotification("buildTarget/didChange")
-    void onBuildTargetChanged(DidChangeBuildTarget params);
+    void onBuildTargetDidChange(DidChangeBuildTarget params);
 
     @JsonRequest("build/registerFileWatcher")
-    CompletableFuture<RegisterFileWatcherResult> registerFileWatcher(RegisterFileWatcherParams params);
+    CompletableFuture<RegisterFileWatcherResult> buildRegisterFileWatcher(RegisterFileWatcherParams params);
 
     @JsonRequest("build/cancelFileWatcher")
-    CompletableFuture<CancelFileWatcherResult> cancelFileWatcher(CancelFileWatcherParams params);
+    CompletableFuture<CancelFileWatcherResult> buildCancelFileWatcher(CancelFileWatcherParams params);
 
     @JsonNotification("buildTarget/compileReport")
-    void onCompileReport(CompileReport params);
+    void onBuildTargetCompileReport(CompileReport params);
 
     default void onConnectWithServer(BuildServer server) {
 
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java
@@ -1,0 +1,35 @@
+package ch.epfl.scala.bsp4j;
+
+import org.eclipse.lsp4j.jsonrpc.services.JsonNotification;
+import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
+
+import java.util.concurrent.CompletableFuture;
+
+public interface BuildClient {
+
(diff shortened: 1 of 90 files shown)


bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java
```diff
@@ -0,0 +1,35 @@
+package ch.epfl.scala.bsp4j;
+
+import org.eclipse.lsp4j.jsonrpc.services.JsonNotification;
+import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
+
+import java.util.concurrent.CompletableFuture;
+
+public interface BuildClient {
```
olafurpg: The method names here are up for debate, please propose ideas 😊

bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildServer.java
```diff
@@ -0,0 +1,53 @@
+package ch.epfl.scala.bsp4j;
+
+import org.eclipse.lsp4j.jsonrpc.services.JsonNotification;
+import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
+
+import java.util.concurrent.CompletableFuture;
+
+public interface BuildServer {
```
olafurpg: The method names here are up for debate, please propose ideas 😊

bsp4j/src/test/scala/tests/BloopSuite.scala
```diff
@@ -0,0 +1,237 @@
...
+  override def publishDiagnostics(params: PublishDiagnosticsParams): Unit = diagnostics += params
+  override def didChangeBuildTarget(params: DidChangeBuildTarget): Unit = pprint.log(params)
+  override def registerFileWatcher(
+      params: RegisterFileWatcherParams): CompletableFuture[RegisterFileWatcherResult] = null
+  override def cancelFileWatcher(
+      params: CancelFileWatcherParams): CompletableFuture[CancelFileWatcherResult] = null
+  override def compileReport(params: CompileReport): Unit = compileReports += params
+}
+
+class BloopSuite extends FunSuite {
```
olafurpg: Example end-to-end usage of this API

bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java
```diff
@@ -0,0 +1,35 @@
...
+    void showMessage(ShowMessageParams params);
+
+    @JsonNotification("build/logMessage")
+    void logMessage(LogMessageParams params);
+
+    @JsonNotification("build/publishDiagnostics")
+    void publishDiagnostics(PublishDiagnosticsParams params);
+
+    @JsonNotification("buildTarget/didChange")
+    void didChangeBuildTarget(DidChangeBuildTarget params);
```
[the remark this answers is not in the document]
olafurpg: Looking at the interface in isolation I agree that `buildTargetDidChange` makes sense but from the usage of the interface I think we can take liberty and use more idiomatic java names

Notifications start with `onX`
- s/buildTargetDidChange/onBuildTargetChanged/
- s/logMessage/onLogMessage
- s/exit/onExit

Requests **request** something with a verb

- s/buildTargetCompile/compileBuildTarget/
- s/workspaceBuildTargets/listWorkspaceBuildTargets/
- s/buildTargetRun/runBuildTarget/
- s/buildTargetTextDocuments/listBuildTargetTextDocuments/
- s/textDocumentBuildTargets/listTextDocumentBuildTargets/


I won't die on a hill defending this approach but my so far limited experience writing the client makes me prefer to write this way

```
val buildServer = connectToBuildServer(...)
// before
buildServer.buildTargetCompile(...)
buildServer.buildTargetRun(...)
buildServer.workspaceBuildTargets(...)
// after
buildServer.compileBuildTarget(...)
buildServer.runBuildTarget(...)
buildServer.listWorkspaceBuildTargets(...)

File: bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java [hunk 5]
```diff
@@ -0,0 +1,35 @@
...
+
+    @JsonRequest("build/registerFileWatcher")
+    CompletableFuture<RegisterFileWatcherResult> registerFileWatcher(RegisterFileWatcherParams params);
+
+    @JsonRequest("build/cancelFileWatcher")
+    CompletableFuture<CancelFileWatcherResult> cancelFileWatcher(CancelFileWatcherParams params);
+
+    @JsonNotification("buildTarget/compileReport")
+    void compileReport(CompileReport params);
```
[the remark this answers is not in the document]
olafurpg: lsp4j uses a feature called "segments" so that all `buildTarget/*` methods are defined in shared trait like `BuildTargetService`

```scala
trait BuildServer {
  def getWorkspaceService: WorkspaceService
  def getBuildService: BuildService
  def getBuildTargetService: BuildTargetService
}
trait WorkspaceService {
  def buildTarget()
}
trait BuildService {
  def registerFileWatcher()
}
trait BuildTargetService {
  def compileReport(..)
}
```

Although this pleases my OCD by being consistent with the protocol I found it awkward to use in practice compared to a flat list of methods.

olafurpg:
I added some details in the commit message of https://github.com/scalacenter/bsp/commit/aba370f22a40d61cdfeadef1fd6428e1b8a4da65

So far I'm enjoying quite a bit the property that all notifications start with `onX`, it helps readability.

jastice:
I agree `onX` is a good idea for notifications. I'm not super emotional about names, but I would much prefer requests also follow a 1:1 correspondence with naming in protocol rather than have English grammar mangle them. Direct correspondence makes it much easier to cross-reference with specification and translate from implementations in other libraries, even if it's a little less natural to write directly.

bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java
```diff
@@ -0,0 +1,35 @@
...
+public interface BuildClient {
+
+    @JsonNotification("build/showMessage")
+    void onBuildShowMessage(ShowMessageParams params);
+
+    @JsonNotification("build/logMessage")
+    void onBuildLogMessage(LogMessageParams params);
+
+    @JsonNotification("build/publishDiagnostics")
+    void onBuildPublishDiagnostics(PublishDiagnosticsParams params);
```
jvican: Maybe we can still satisfy @jastice's request of keeping the same names as the spec by using the real event name but preprended by `on*` like `onPublishDiagnosticsParams`?

olafurpg: the current `onBuildPublishDiagnostics` name matches the protocol method name instead of the paramter type.

docs/bsp.md
```diff
@@ -1194,7 +1194,7 @@ trait ScalaMainClassesItem {
   def target: BuildTargetIdentifier
 
   /** The main class item. */
-  def mainClass: ScalaMainClass
+  def classes: List[ScalaMainClass]
```
jvican: Why this change? A `ScalaMainClassItem` links a class to a build target identifier, and the result returns a list of `ScalaMainClassItem`s.

build.sbt
```diff
@@ -10,3 +21,48 @@ val bsp = project
...
+      IO.delete(outDir)
+      compiler.setOutputPath(outDir.getCanonicalPath)
+      object XtendError
+          extends Exception(s"Compilation of Xtend files in $sourceDir failed.")
+          with sbt.internal.util.FeedbackProvidedException
+      if (!compiler.compile())
+        throw XtendError
+    },
+    unmanagedSourceDirectories.in(Compile) += sourceDirectory.in(Compile).value / "xtend-gen",
+    resourceGenerators.in(Test) += Def.task {
```
jvican: Resource generators are run on every `compile` invocation, even if you only compile a file. I don't think it's necessary for this PR but it would be better if we only codegen when the xtend sources have changed.

olafurpg: The xtend code generation needs to be triggered manually via `sbt bsp4j/codegen` whenever an `*.xtend` file is changed. This is a resource generator for the tests, it only writes the working directory of the bloop test directory, which I expect is cheap.

docs/bsp.md
olafurpg: This is synchronizing the spec with the current bsp4s implementation 
https://github.com/scalacenter/bsp/blob/49d840991f5e57ef3af1e535328c2baa179c1d75/src/main/scala/ch/epfl/scala/bsp/Bsp.scala#L543

PR #43, opened 2018-10-25
Remove file watcher capabilities

jvican:
I added file watching capabilities in https://github.com/scalacenter/bsp/pull/13. The design of the APIs was
done jointly with Olaf, where we itereated on several solutions until
we arrived to this one.

I think the design of the feature was great and served a good purpose:
allowing tools like IntelliJ to communicate more efficiently with BSP
servers and tell the build tool what things changed or not in the editor
side.

However, I believe it was too soon to add it to the protocol and it's
better if we remove it now for BSP 2.0 and add it back if we find
ourselves needing it. So far, no client or server had yet implemented
file watching capabilities in BSP, so it's fine to remove it.


jvican:
Thanks for the reviews guys, CI is passing so I'm merging.

PR #44, opened 2018-10-25
Add status code to compile and test

### LINKED ISSUE #41: Add status code to compile and test result

jvican:
Replaces the previous mechanism, whereby build tools implementing BSP
would use a BSP protocol error to signal failure in compilation or
testing. Now, this new mechanism will simplify the handling on the
client side.
  
Fixes #41.


jvican:
Addressed.

PR #45, opened 2018-10-25
Add buildTarget/cleanCache request (fix #34)

### LINKED ISSUE #34: Run task without cache vs cleaning
We want to support a way for clients to run tasks, especially compiles, without cache, i.e. a non-incremental build.

One way to do this is to support a `clean` request that asks to remove any products of a build target.

Another way would be to add a flag to the compile request, and potentially other tasks, that requests a full rebuild.

What considerations affect the choice of these options, and are there other options?

olafurpg:


docs/bsp.md
```diff
@@ -913,6 +913,33 @@ during compilation before completing the response.
 The client will get a `originId` field in `RunResult` if the `originId` field in the
 `RunParams` is defined.
 
+### Clean Cache Request
+
+The clean cache request is sent from the client to the server to remove cached artifacts that are associated with a given build target.
```
jastice: I think some clarification of semantics would be good here, with the acknowledgement that they may be tool-specific.
For example:
After a successful cleanCache request, it is expected that disk artifacts created by tasks on the target have been removed, and compile requests to the given targets will perform a full recompile. The build tool defines the exact semantics.

olafurpg: Good point, done.

docs/bsp.md
```diff
@@ -913,6 +914,36 @@ during compilation before completing the response.
...
+An example use-case for the clean cache request is to free up disk space or troubleshoot build problems.
+
+After a successful clean cache request, it is expected that cached artifacts created by invoking other requests (compile, run, test) on the build target have been removed and subsequent requests result in a full recompilation.
```
jvican: Can we tone down semantics of the `clean` operation in this paragraph? In particular, the part about "cached artifacts" (which sound like cached files on a hard drive) and the previous part about example use cases of `clean`.
  
I'm afraid the current wording can be misread by people to mean the client can expect "free disk space" to be a side effect of the operation, which makes it look like the server is forced to clean caches on disk. In bloop, a `clean` operation cleans only the incremental compiler state by passing a fresh analysis to `compile`, so we could not comply with the current wording.
  
What do you think about changing the description of this request with the following paragraph? I've tried to remove the part of example use cases and be more clear about what a user can expect (in particular, I liked the way you put it with the full recompilation expectation).
  
> The clean cache request is sent from the client to the server to reset any state associated with a given build target. The state can live either in the build tool or in the file system.
>
> The build tool defines the exact semantics of the clean cache request:
>   1. Stateless build tools are free to ignore the request and respond with a successful response.
>   2. Stateful build tools must ensure that invoking compilation on a target that has been cleaned results in a full compilation.

olafurpg: That is a good formulation, I updated the description to use it unchanged.

olafurpg:
I'll go ahead and merge this 💪

olafurpg:
Sorry @jastice just realized you hadn't approved yet. I'm happy to follow up if you have further comments.

jastice:
It's fine with me. Personally I think the name `clean` would be a better fit, but that's bikeshedding

PR #47, opened 2018-10-25
Add buildTarget/sources request and remove buildTarget/textDocuments 

olafurpg:


docs/bsp.md
```diff
@@ -578,27 +579,51 @@ object BuildTargetEventKind {
...
+  def kind: Int
+}
+object SourceItemKind {
+  /** Source code that is manually edited by the user. */
+  val Source = 1
+  /** Data files that are manually managed by the user. */
+  val Resource = 2
+  /** Source code that is automatically generated by the build and is not
+    * intended to be manually edited by the user. */
+  val Excluded = 3
```
jastice: I think `Generated` is a better name here. Also it's orthogonal to source/resource. Perhaps have a `generated` flag on `SourceItem` instead?

docs/bsp.md
```diff
@@ -578,27 +579,51 @@ object BuildTargetEventKind {
...
+  def uri: Uri
+
+  /** The category of this source item, must be one of SourceItemKind. */
+  def kind: Int
+}
+object SourceItemKind {
+  /** Source code that is manually edited by the user. */
+  val Source = 1
+  /** Data files that are manually managed by the user. */
+  val Resource = 2
```
jvican: :thinking: Why do we need this kind if we already have `buildTarget/resources` to get the resources of a build target?
  
I suggest we remove `Resource` and use rename `Source` to `Default` and `Excluded` to `Generated` (@jastice seems to propose the same thing in a comment below). What do you think?

docs/bsp.md
olafurpg: I propose we

- remove `SourceItemKind`
- add `SourceItem.generated: Boolean`

As @jvican pointed out, we already have `buildTarget/resources.

docs/bsp.md
olafurpg: Good point, see https://github.com/scalacenter/bsp/pull/47#discussion_r228447419

olafurpg:
Rebased on master, waiting approval from @jastice

Issue #7, opened 2018-04-13
Add entrypoint for compile progress notifications


jvican:
Build tools have ways to hook into compilers to get progress notifications. For example, this is Zinc's interface:
  
```scala
/*
 * Zinc - The incremental compiler for Scala.
 * Copyright 2011 - 2017, Lightbend, Inc.
 * Copyright 2008 - 2010, Mark Harrah
 * This software is released under the terms written in LICENSE.
 */

package xsbti.compile;

/**
 * An API for reporting when files are being compiled.
 *
 * This design is tied to the Scala compiler but it is used
 * by the Java compiler too, check the docs of the methods.
 */
public interface CompileProgress {

	/**
	 * Start the progress of a concrete phase for the path of a given
	 * compilation unit.
	 *
	 * @param phase The phase of the compiler being run.
	 * @param unitPath The path of the compilation unit. It will be empty
	 *                 when a Java compiler is reporting the progress.
	 */
	void startUnit(String phase, String unitPath);

	/**
	 * Advance the progress of the current unit.
	 *
	 * @param current The current progress.
	 * @param total The total of the progress that has to be achieved.
	 *
	 * @return Whether the user has cancelled compilation or not.
	 */
	boolean advance(int current, int total);
}
```
  
Progress notifications are useful to notify users of IDEs and editors how is compilation going, and have a feeling of what are the slow phases (or even the slow compilation units) to compile. The best way to know that you have a slow project to compile is to expose this information to the users!
  
This ticket registers my intention of adding such an entrypoint in BSP, under the name of `buildTarget/showCompileProgress`. I haven't yet figured out the right interface, but it would look similar to Zinc's interface. We may even need to define two events to support every method, and then an event kind to the `ShowCompileProgressParams`.
  
This entrypoint would allow IntelliJ, for example, to mimic the same user experience than the built-in "Build Project" which shows compilation progress notifications in a small event window at the bottom right.

jastice:
I think the parameters need to have some kind of threading id to model the tree structure. Something like:
```
startUnit(String parentId, String id, String phase, String unitPath)
advance(String id, int current, int total)
```

jastice:
For reference, the progress API in the IntelliJ build tool window looks like this:
```java
public StartEventImpl(
    @NotNull Object eventId, 
    @Nullable Object parentId, 
    long eventTime, 
    @NotNull String message)

public StartEventImpl(
    @NotNull Object eventId, 
    @Nullable Object parentId, 
[...]

PR #49, opened 2018-10-25
Describe task notifications and how they are used with requests

jastice:
Fleshing out the proposal in #7, describe a spec for how task progress notifications work, and how they are used to communicate status in compile and test requests.

### DIFF
--- a/bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
+++ b/bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
@@ -3,8 +3,8 @@ package ch.epfl.scala.bsp
 import java.net.{URI, URISyntaxException}
 
 import io.circe.Decoder.Result
-import io.circe.{Decoder, DecodingFailure, HCursor, Json, RootEncoder}
 import io.circe.derivation.JsonCodec
+import io.circe._
 
 final case class Uri private[Uri] (val value: String) {
   def toPath: java.nio.file.Path =
@@ -332,6 +332,10 @@ case object BuildTargetEventKind {
     time: Option[Long]
 )
 
+@JsonCodec final case class CompileTask(
+    target: BuildTargetIdentifier
+)
+
 @JsonCodec final case class TestParams(
     targets: List[BuildTargetIdentifier],
     originId: Option[String],
@@ -355,6 +359,50 @@ case object BuildTargetEventKind {
     pending: Int,
     time: Option[Long]
 )
+
+@JsonCodec final case class TestTask(
+    target: BuildTargetIdentifier
+)
+
+@JsonCodec final case class TestStart(
+    description: String,
+    location: Option[Location]
+)
+
+@JsonCodec final case class TestFinish(
+    description: String,
+    message: Option[String],
+    status: TestStatus,
+    location: Option[Location],
+    data: Option[Json]
+)
+
+
+sealed abstract class TestStatus(val code: Int)
+object TestStatus {
+  case object Passed extends TestStatus(1)
+  case object Failed extends TestStatus(2)
+  case object Ignored extends TestStatus(3)
+  case object Cancelled extends TestStatus(4)
+  case object Skipped extends TestStatus(5)
+
+  implicit val testStatusEncoder: RootEncoder[TestStatus] = new RootEncoder[TestStatus] {
+    override def apply(a: TestStatus): Json = Json.fromInt(a.code)
+  }
+
+  implicit val statusCodeDecoder: Decoder[TestStatus] = new Decoder[TestStatus] {
+    override def apply(c: HCursor): Result[TestStatus] = {
+      c.as[Int].flatMap {
+        case 1 => Right(Passed)
+        case 2 => Right(Failed)
+        case 3 => Right(Ignored)
+        case 4 => Right(Cancelled)
+        case 5 => Right(Skipped)
+        case n => Left(DecodingFailure(s"Unknown test status $n", c.history))
+      }
+    }
+  }
+}
 
 @JsonCodec final case class RunParams(
     target: BuildTargetIdentifier,
@@ -397,6 +445,55 @@ object StatusCode {
     message: Option[String],
     cleaned: Boolean
 )
+
+@JsonCodec final case class TaskStartParams(
+    taskId: TaskId,
+    eventTime: Option[Long],
+    message: Option[String],
+    dataKind: Option[String],
+    data: Option[Json]
+)
+
+@JsonCodec final case class TaskProgressParams(
+    taskId: TaskId,
+    eventTime: Option[Long],
+    message: Option[String],
+    total: Option[Long],
+    progress: Option[Long],
+    unit: Option[String],
+    dataKind: Option[String],
+    data: Option[Json]
+)
+
+@JsonCodec final case class TaskFinishParams(
+    taskId: TaskId,
+    eventTime: Option[Long],
+    message: Option[String],
+    status: StatusCode,
+    dataKind: Option[String],
+    data: Option[Json]
+)
+
+object DataKind {
+
+  /** `data` field must contain a CompileTask object. */
+  val CompileTask = "compile-task"
+
+  /** `data` field must contain a CompileReport object. */
+  val CompileReport = "compile-report"
+
+  /** `data` field must contain a TestTask object. */
+  val TestTask = "test-task"
+
+  /** `data` field must contain a TestReport object. */
+  val TestReport = "test-report"
+
+  /** `data` field must contain a TestStart object. */
+  val TestStart = "test-start"
+
+  /** `data` field must contain a TestFinish object. */
+  val TestFinish = "test-finish"
+}
 
 sealed abstract class ScalaPlatform(val id: Int)
 object ScalaPlatform {
--- a/bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
+++ b/bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
@@ -474,24 +474,12 @@ object StatusCode {
     data: Option[Json]
 )
 
-object DataKind {
-
-  /** `data` field must contain a CompileTask object. */
+object TaskDataKind {
   val CompileTask = "compile-task"
-
-  /** `data` field must contain a CompileReport object. */
   val CompileReport = "compile-report"
-
-  /** `data` field must contain a TestTask object. */
   val TestTask = "test-task"
-
-  /** `data` field must contain a TestReport object. */
   val TestReport = "test-report"
-
-  /** `data` field must contain a TestStart object. */
   val TestStart = "test-start"
-
-  /** `data` field must contain a TestFinish object. */
   val TestFinish = "test-finish"
 }
 
--- a/docs/bsp.md
+++ b/docs/bsp.md
@@ -795,10 +795,10 @@ trait TaskStartParams {
     def taskId: TaskId
 
     /** Timestamp of when the event started in milliseconds since Epoch. */
-    def eventTime: Long
+    def eventTime: Option[Long]
 
     /** Message describing the task. */
-    def message: String
+    def message: Option[String]
     
     /** Kind of data to expect in the `data` field. If this field is not set, the kind of data is not specified.
       * Kind names for specific tasks like compile, test, etc are specified in the protocol.
@@ -825,10 +825,10 @@ trait TaskProgressParams {
     def taskId: TaskId
 
     /** Timestamp of when the progress event was generated in milliseconds since Epoch. */
-    def eventTime: Long
+    def eventTime: Option[Long]
 
     /** Message describing the task. */
-    def message: String
+    def message: Option[String]
 
     /** If known, total amount of work units in this task. */
     def total: Option[Long]
@@ -837,7 +837,7 @@ trait TaskProgressParams {
     progress: Option[Long]
 
     /** Name of a work unit. For example, "files" or "tests". May be empty. */
-    unit: String
+    unit: Option[String]
     
     /** Kind of data to expect in the `data` field. If this field is not set, the kind of data is not specified.
       * Kind names for specific tasks like compile, test, etc are specified in the protocol.
@@ -863,10 +863,10 @@ trait TaskFinishParams {
     def taskId: TaskId
 
     /** Timestamp of the event in milliseconds. */
-    def eventTime: Long
+    def eventTime: Option[Long]
 
     /** Message describing the finish event. */
-    def message: String
+    def message: Option[String]
 
     /** Task completion status. */
     def status: StatusCode
@@ -1103,8 +1103,8 @@ trait TestFinish {
   /** Name or description of the test. */
   def description: String
 
-  /** Information about completion of the test, for example error message. */
-  def message: String
+  /** Information about completion of the test, for example an error message. */
+  def message: Option[String]
 
   /** Completion status of the test. */
   def status: Int
--- a/docs/bsp.md
+++ b/docs/bsp.md
@@ -1088,10 +1088,10 @@ This request may trigger a compilation or other tasks on the selected build targ
 The server may send any number of `build/task*`, `build/publishDiagnostics` and `build/logMessage` notifications
 to communicate about tasks triggered by the request to the client.
 
-#### Test Unit Notifications
+#### Test Notifications
 
-The server may inform about individual test units in task notifications that reference the originating task in their `taskId`.
-For example, the server can send a `taskStart`/`taskFinish` for each test suite in a target, and likewise for each test in the suite.
+The server may inform about individual tests or groups of tests in task notifications that reference the originating task in their `taskId`.
+For example, the server can send a `taskStart`/`taskFinish` for each test suite in a target, and likewise for each individual test in the suite.
 The server's implementation decides the granularity at which tests are reported. For example, if it only has information 
 about all the tests in a suite at a time, it could report a TestFinish for each test once the suite is done.
 
--- a/docs/bsp.md
+++ b/docs/bsp.md
@@ -791,10 +791,10 @@ Notification:
 
 ```scala
 trait TaskStartParams {
-    /** Unique id of the task with reference to parent task id */
+    /** Unique id of the task with optional reference to parent task id */
     def taskId: TaskId
 
-    /** Timestamp of the event in milliseconds. */
+    /** Timestamp of when the event started in milliseconds since Epoch. */
     def eventTime: Long
 
     /** Message describing the task. */
@@ -803,7 +803,7 @@ trait TaskStartParams {
     /** Optional metadata about the task.
       * For compilation tasks, this should be a CompileTask object.
       * For test tasks, this should be a TestTask object. */
-    def data: Option[Json]
+    def data: Option[TaskData]
 }
 ```
 
@@ -816,10 +816,10 @@ After a `taskStart` and before `taskFinished` for a `taskId`, the server may sen
 
 ```scala
 trait TaskProgressParams {
-    /** Unique id of the task with reference to parent task id */
+    /** Unique id of the task with optional reference to parent task id */
     def taskId: TaskId
 
-    /** Timestamp of the event in milliseconds. */
+    /** Timestamp of when the progress event was generated in milliseconds since Epoch. */
     def eventTime: Long
 
     /** Message describing the task. */
@@ -835,7 +835,7 @@ trait TaskProgressParams {
     unit: String
 
     /** Optional metadata about the task progress. */
-    data: Option[Json]
+    data: Option[TaskData]
 }
 
 ```
@@ -847,13 +847,13 @@ trait TaskProgressParams {
 
 ```scala
 trait TaskFinishParams {
-    /** Unique id of the task with reference to parent task id */
+    /** Unique id of the task with optional reference to parent task id */
     def taskId: TaskId
 
     /** Timestamp of the event in milliseconds. */
     def eventTime: Long
 
-    /** Message describing the task. */
+    /** Message describing the finish event. */
     def message: String
 
     /** Task completion status. */
@@ -862,9 +862,24 @@ trait TaskFinishParams {
     /** Optional metadata about the task result.
       * For compilation tasks, this should be a CompileReport object.
       * For test tasks, this should be a TestReport object. */
-    def data: Option[Json]
+    def report: Option[TaskData]
 }
 ```
+
+#### Task Data
+
+Task progress notifications may contain a `TaskData` object in their `data` field.
+There are predefined kinds of `TaskData` for test and compile tasks, as defined in the [Compile Request](#compile-request)
+and [Test Request](#test-request) sections.
+
+```scala
+trait TaskData {
+  /** Identifier for the kind of task data encoded in the `data` field. */
+  def kind: String
+  def data: Json
+}
+```
+
 
 ### Compile Request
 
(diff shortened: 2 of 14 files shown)


docs/bsp.md
```diff
@@ -697,6 +716,106 @@ trait ResourcesItem {
...
+* method: `build/taskStart`
+* params: `TaskStartParams` defined as follows:
+
+```scala
+trait TaskStartParams {
+    /** Unique id of the task with reference to parent task id */
+    def taskId: TaskId
+
+    /** Timestamp of the event in milliseconds. */
+    def eventTime: Long
```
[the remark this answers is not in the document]
olafurpg: How would the client know how long an ongoing target has been running for? For example, to display messages like: "Downloading sources (started 34 seconds ago)"

I think it's good to keep `eventTime` to match the API proposed ini https://github.com/scalacenter/bsp/issues/7#issuecomment-431829529

docs/bsp.md
```diff
@@ -697,6 +716,106 @@ trait ResourcesItem {
...
+
+* method: `build/taskStart`
+* params: `TaskStartParams` defined as follows:
+
+```scala
+trait TaskStartParams {
+    /** Unique id of the task with reference to parent task id */
+    def taskId: TaskId
+
+    /** Timestamp of the event in milliseconds. */
```
olafurpg: suggestion:

> Timestamp of when the event started in milliseconds.

docs/bsp.md
```diff
@@ -697,6 +716,106 @@ trait ResourcesItem {
...
+#### Task Progress
+
+After a `taskStart` and before `taskFinished` for a `taskId`, the server may send any number of progress notifications.
+
+* method: `build/taskProgress`
+* params: `TaskProgressParams` defined as follows:
+
+```scala
+trait TaskProgressParams {
+    /** Unique id of the task with reference to parent task id */
```
olafurpg: > s/with reference to parent task id//

The parent id is optional

docs/bsp.md
```diff
@@ -697,6 +716,106 @@ trait ResourcesItem {
...
+    def taskId: TaskId
+
+    /** Timestamp of the event in milliseconds. */
+    def eventTime: Long
+
+    /** Message describing the task. */
+    def message: String
+
+    /** If known, total amount of work units in this task. */
+    def total: Option[Long]
```
olafurpg: How about making this `Double` instead of `Long`?

docs/bsp.md
```diff
@@ -697,6 +716,106 @@ trait ResourcesItem {
...
+
+```scala
+trait TaskFinishParams {
+    /** Unique id of the task with reference to parent task id */
+    def taskId: TaskId
+
+    /** Timestamp of the event in milliseconds. */
+    def eventTime: Long
+
+    /** Message describing the task. */
```
olafurpg: > Message describing the finish event, can be different from progress message.

docs/bsp.md
```diff
@@ -697,6 +716,106 @@ trait ResourcesItem {
...
+
+```scala
+trait TaskProgressParams {
+    /** Unique id of the task with reference to parent task id */
+    def taskId: TaskId
+
+    /** Timestamp of the event in milliseconds. */
+    def eventTime: Long
+
+    /** Message describing the task. */
```
olafurpg: > Message describing the progress event, can be different from the start message and can be different from other progress events with the same task id.

docs/bsp.md
```diff
@@ -697,6 +716,106 @@ trait ResourcesItem {
...
+    def total: Option[Long]
+
+    /** If known, completed amount of work units in this task. */
+    progress: Option[Long]
+
+    /** Name of a work unit. For example, "files" or "tests". May be empty. */
+    unit: String
+
+    /** Optional metadata about the task progress. */
+    data: Option[Json]
```
olafurpg: Instead of `data: Option[Json]`, how about adding optional fields for test and compile reports?

```scala
def compileReport: Option[CompileReport]
def testReport: Option[TestReport]
```

That would simplify the serialization/deserialization.

docs/bsp.md
```diff
@@ -857,14 +975,54 @@ trait TestReport {
...
+The server may send any number of `build/task*`, `build/publishDiagnostics` and `build/logMessage` notifications
+to communicate about tasks triggered by the request to the client.
+
+Subtask notifications:
+
+The server may inform about individual test units in subtask notifications. Individual test notifications should
+include the `TestStarted` and `TestFinished` objects in their params:
+
+```scala
+trait TestStarted {
```
olafurpg: Great idea! This was missing in the previous protocol 👍

docs/bsp.md
```diff
@@ -857,14 +975,54 @@ trait TestReport {
...
+    val Failed: Int = 2
+
+    /** The test was marked as ignored. */
+    val Ignored: Int = 3
+
+    /** The test execution was cancelled. */
+    val Cancelled: Int = 4
+
+    /** The was not included in execution. */
+    val skipped: Int = 5
```
olafurpg: `s/skipped/Skipped/`

docs/bsp.md
```diff
@@ -857,14 +975,54 @@ trait TestReport {
...
+
+Subtask notifications:
+
+The server may inform about individual test units in subtask notifications. Individual test notifications should
+include the `TestStarted` and `TestFinished` objects in their params:
+
+```scala
+trait TestStarted {
+  /** Name of description of the test. */
+  def description: String
```
olafurpg: Should this be called `name` instead? Would be good to explicitly mention if it must/should be unique or not.

docs/bsp.md
```diff
@@ -857,14 +975,54 @@ trait TestReport {
...
-This request may trigger a compilation on the selected build targets. The server is free to send any
-number of `build/compileReport`, `build/publishDiagnostics` and `build/logMessage` notifications
-during compilation before completing the response.
+The server may send any number of `build/task*`, `build/publishDiagnostics` and `build/logMessage` notifications
+to communicate about tasks triggered by the request to the client.
+
+Subtask notifications:
+
+The server may inform about individual test units in subtask notifications. Individual test notifications should
+include the `TestStarted` and `TestFinished` objects in their params:
```
olafurpg: Can you elaborate on "include the `TestStarted` and `TestFinished` objects in their params:"?

docs/bsp.md
```diff
@@ -824,19 +936,25 @@ trait TestResult {
 
 The field `data` may contain test-related language-specific information.
 
-Notification:
+Notifications:
+
+The beginning of a testing unit may be signalled to the client with a `build/taskStart` notification.
+For testing of a build target, the notification params should include a `TestTask` object:
```
olafurpg: Can you clarify "the notification params should include a `TestTask` object:"?

docs/bsp.md
```diff
@@ -777,11 +892,8 @@ trait CompileReport {
 ```
 
 The server is free to send any number of `build/publishDiagnostics` and `build/logMessage`
-notifications during compilation before completing the response. The client is free to forward these
-messages to the LSP editor client.
-
-The client will get a `originId` field in `CompileReport` or `CompileResult` if the `originId`
-field in the `CompileParams` is defined.
+notifications during compilation before completing the response. Any number of subtasks may be communicated
```
olafurpg: Can you clarify what "subtask" means? It's mentioned three times but not explicitly introduced.

docs/bsp.md
```diff
@@ -739,23 +858,19 @@ trait CompileResult {
 }
 ```
 
-where `StatusCode` is defined as follows
+Notifications:
+
+The beginning of a compilation unit may be signalled to the client with a `build/taskStart` notification.
+For compilation of a build target, the notification params should include `CompileTask` object:
```
olafurpg: Same question as for `TestTask`, can you clarify "For compilation of a build target, the notification params should include `CompileTask` object:"?

olafurpg:
There's related discussion in https://github.com/scalacenter/bsp/issues/7#issuecomment-433381538

Let's try to keep the discussion here in the PR instead.

docs/bsp.md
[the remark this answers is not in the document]
olafurpg: Here's one example

screen shot 2018-10-26 at 13 49 44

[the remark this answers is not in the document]
jastice: Yup, this is exactly the api that is supported by IntelliJ. Afaik mill also shows task running times and Eugene was working on something like this for sbt.

docs/bsp.md
jastice: What use case do you have in mind? Doubles can get messy (admittedly not in most normal circumstances) and in builds we usually deal with discrete units.

olafurpg: I'm fine keeping `Long`. My thinking was that JSON numbers are doubles at the end of the day and progress bars are normally represented as percentages as well, but that doesn't prohibit us from still  using `Long`.

docs/bsp.md
jastice: Was supposed to be "name *or* description". It's intended to be descriptive :) Not necessarily unique. Should probably add a source location parameter for the test.

docs/bsp.md
jastice: It is supposed to mean the `data` parameter. I'll add that.

docs/bsp.md
jastice: It is supposed to mean the `data` parameter.

jvican:
Instead of adding an optional field for the core tasks, I would introduce a report API and kinds. I think this approach is cleaner and will avoid polluting the task progress API every time we want to add a core data structure.
  

```diff
trait TaskFinishParams {
-    /** Optional metadata about the task.
-      * For compilation tasks, this should be a CompileTask object.
-      * For test tasks, this should be a TestTask object. */
-    def data: Option[Json]
+    def kind: ReportKind
+    def data: Option[Report]
}
+
+ object ReportKind {
+    val Generic = 0
+    val Compilation = 1
+    val Test = 2
+}
+  
+trait Report {
+    def data: Json
+}
```
  
and then modify the spec to tell clients to serialize a `CompileReport` when the `Compilation` kind is attached to a task start params and viceversa for `TestReport`, with a fallback on the generic report if the kind is unknown.

jastice:
> Instead of adding an optional field for the core tasks, I would introduce a report API and kinds. I think this approach is cleaner and will avoid polluting the task progress API every time we want to add a core data structure.

That looks like a good approach to me. It allows using the generic task notifications without making the client guess what kind of task it is dealing with.

olafurpg:
I like that approach @jvican 

Instead of limiting the number `ReportKind` we can make it open for extensions with `kind: Option[String]` with predefined values
```scala
object ReportKind {
  val Compile = "compile"
  val Test = "test"
  val Run = "run" 
}
```

That avoids the `Default = 1` kind.

olafurpg:
I am not sure we need `report: Option[Report]`, we already use `data: Option[Json]` in other endpoints. Alternatively, we can make it

```scala
trait TaskProgress {
  def report: Option[TaskReport]
}
trait TaskReport {
  def kind: String
  def data: Option[Json]
}
```

I don't have a strong opinion either way it goes, I'm happy with the overall idea.

docs/bsp.md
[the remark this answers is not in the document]
jastice: potential problem: A remote server has a totally different clock than local client. But I think this is rare and not fatal.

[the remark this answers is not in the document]
olafurpg: As long as the timestamps are created by the build server it should be a non-problem.

jastice:
I've updated with clarifications and defined the `TaskData` wrapper object that contains the tag how to interpret the `data` object. I'm not sure I like it because it requires wrapping/unwrapping an additional object to get at the data.

docs/bsp.md
olafurpg: .

docs/bsp.md
```diff
@@ -830,10 +830,25 @@ trait TaskFinishParams {
     /** Optional metadata about the task result.
       * For compilation tasks, this should be a CompileReport object.
       * For test tasks, this should be a TestReport object. */
-    def data: Option[Json]
+    def report: Option[TaskData]
```
olafurpg: How about keeping `data: Option[Json]` unchanged and inlining `kind: Option[String]`? That avoids `TaskData` altogether
```scala
  def dataKind: Option[String]
    /** Optional metadata about the task result.
      * For compilation tasks, this should be a CompileReport object.
      * For test tasks, this should be a TestReport object. */
    def data: Option[Json]
}
object TaskDataKind {
  val CompileReport = "compile-report"
  // ...
} 
```

jastice: I think I like this better afterall, yes. But maybe with one less optional:
```
/** The kind of the `data` field. "NONE" if `data` is empty. */
def dataKind: String
def data: Option[Json]

object DataKind {
  val NONE = "NONE"
  val CompileReport = "compile-report"
  // ...
}
```

olafurpg: I think we should use what makes more sense in JSON which is that the `dataKind` key is either null or undefined, in bsp4j it will be a `String` either way.

olafurpg: It's `Option[String]` in bsp4s to support the case when the key is missing, I think we actually have too many non-optional fields

jastice:
Feature creep: I'd like a way to show/receive tasks based on some detail level, similar to log levels.
For example, if sbt sends me a task notification for every sbt task, that would get very spammy in general case, but be very convenient for debugging and figuring out bottlenecks.

What do you think of a `level: Int` field in `TaskStarted` / `TaskFinished` notifications that signifies how important a task is?

Or should we just leave this part to build tool configuration?

docs/bsp.md
jastice: What do you mean? When it's not optional on the porotocol level it's simply a protocol error when it's not there, for optional it's up to the client to figure things out.

But I guess in this case an option is just as well.

jastice:
Okay, updated with inlined `dataKind`. Please have a look, if we agree I will implement the library encodings.

docs/bsp.md
```diff
@@ -742,6 +760,155 @@ trait ResourcesItem {
...
+### Task Notifications
+
+The server may inform the client on the state of running tasks, such as compilation or tests.
+When beginning a task, the server may send `build/taskStart`, intermediate updates may be sent in
+`build/taskProgress`.
+
+If a `build/taskStart` notification has been sent, the server must send `build/taskFinish`
+on completion of the same task.
+
+Conversely, a `build/taskFinish` notification may be sent even if `build/taskStart` was not sent.
```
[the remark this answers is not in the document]
jastice: I was thinking of simplifying server implementations here: In some cases they don't really have access to task starts (eg tests in sbt). OTOH, they can just send start and finish together.

docs/bsp.md
```diff
@@ -742,6 +760,155 @@ trait ResourcesItem {
 }
 ```
 
+### Task Notifications
+
+The server may inform the client on the state of running tasks, such as compilation or tests.
```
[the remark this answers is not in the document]
jastice: Do you think it's strictly necessary to send the reports? I'm thinking the simplest implementation will just reply with a failed `CompileResult` etc and let the user read messages for details

docs/bsp.md
```diff
@@ -784,23 +951,21 @@ trait CompileResult {
...
+trait CompileTask {
+  def target: BuildTargetIdentifier
 }
 ```
 
-Notification:
-
-* method: `buildTarget/compileReport`
-* params: `CompileReport` defined as follows:
+The completion of a compilation task should be signalled with a `build/taskFinish` notification.
```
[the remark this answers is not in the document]
jastice: As above, this could be optional (but strongly encouraged)

docs/bsp.md
```diff
@@ -869,19 +1031,27 @@ trait TestResult {
 
 The field `data` may contain test-related language-specific information.
 
-Notification:
+#### Test Notifications
 
-* method: `buildTarget/testReport`
-* params: `TestReport` defined as follows:
+The beginning of a testing unit may be signalled to the client with a `build/taskStart` notification.
```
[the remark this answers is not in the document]
jastice: As with compile, `must` would imply that these notifications have to be sent, but leaving them out should be an option.

docs/bsp.md
```diff
@@ -902,14 +1072,63 @@ trait TestReport {
...
+```scala
+trait TestStart {
+  /** Name or description of the test. */
+  def description: String
+  
+  /** Source location of the test, as LSP location. */
+  def location: Option[Location]
+}
+
+trait TestFinish {
```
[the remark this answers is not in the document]
jastice: I don't think anything about this suggests that these notifications are compulsory. The build tool can implement them at whatever granularity is possible or practical.

For example, if it only has information about all the tests in a suite at a time, it could report a `TestFinish` for each test once the suite is done. If it can report only on the suite as a whole, the same `TestFinish` can be used or just use plain messages with `StatusCode`

bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
```diff
@@ -399,6 +447,55 @@ object StatusCode {
...
+    eventTime: Option[Long],
+    message: Option[String],
+    status: StatusCode,
+    dataKind: Option[String],
+    data: Option[Json]
+)
+
+object DataKind {
+
+  /** `data` field must contain a CompileTask object. */
```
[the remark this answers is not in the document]
jastice: I'll take it out for consistency, but maybe we should?

docs/bsp.md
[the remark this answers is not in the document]
jvican: I think we should make it compulsory to send the reports for compilation and test, yeah. They can be seen as a hitting a "milestone" in compilation and clients can benefit from having it, especially when several modules are compiled for a given compile action.
  
I don't think there's a use case for making them optional, and trivially all build tools can send compilation reports (that's what all of them do whenever they print to console errors and warnings).

docs/bsp.md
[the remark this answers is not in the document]
jvican: I think it's actually easier for build tools to force a finish iff a start notification has been sent. I prefer the principle of this approach and the fact that clients can easily reason about it.

docs/bsp.md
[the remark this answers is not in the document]
jvican: I think it should be a `must` :smile: Let's continue the discussion on the other comment.

docs/bsp.md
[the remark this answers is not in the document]
jvican: > I don't think anything about this suggests that these notifications are compulsory.
  
The documentation of the fields mislead me to think it's compulsory to report on a test-by-test basis, claiming that they are metadata about "tests" only.
  
> For example, if it only has information about all the tests in a suite at a time, it could report a TestFinish for each test once the suite is done. 
  
If this is the case, I propose:
  
1. We explicitly mention that the build tool can use the granularity it wants.
2. To rename the documentation to not be tied to a concrete test.

docs/bsp.md
```diff
@@ -902,14 +1072,63 @@ trait TestReport {
...
+  /** Name or description of the test. */
+  def description: String
```
jvican: Two questions:
  
1. Should this be named `description`? I think we should find a better name.
2. Should this be typed as `String`? Why not use a URI and call it id?

docs/bsp.md
```diff
@@ -869,19 +1031,27 @@ trait TestResult {
...
-Notification:
+#### Test Notifications
 
-* method: `buildTarget/testReport`
-* params: `TestReport` defined as follows:
+The beginning of a testing unit may be signalled to the client with a `build/taskStart` notification.
+When the testing unit is a build target, the notification's `dataKind` field should be `test-task` and the 
+`data` field should include a `TestTask` object.
+
+```scala
```
jvican: I would make stress that a possible use case here for BSP clients is to use the parent-child relationships of task notifications to do a nice tree-rendering of test executions, provided that the build tool respects this hierarchy when sending the task notifications.

docs/bsp.md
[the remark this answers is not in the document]
jastice: I'm thinking for example of a maximally simple bsp wrapper for some command line tool. It would parse exit codes of command line compile invocations and otherwise just forward messages. Or just a really lazy build tool implementor that wants a quick IDE integration while hacking on the tool implementation. I don't think there's a correctness reason to strictly require these notifications.

docs/bsp.md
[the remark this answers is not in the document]
jastice: Agreed

docs/bsp.md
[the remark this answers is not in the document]
jvican: I think we should focus 

> I don't think there's a correctness reason to strictly require these notifications.
  
I agree, I haven't done any correctness reason. What I'm claiming is that:
  
1. This is not a point of friction for build tools. Giving this information is simple and should be forced in the protocol.
2. It makes the life of the IDE clients useful. Clients can (and probably need to) use the notification as a milestone after which they do a concrete action, show a popup or add to a view, etc.
  
I cannot imagine how a maximally simple bsp wrapper would be even able to support the rest of our non-trivial build tool requirements in the protocol, so I think this is not a use case we should support.

docs/bsp.md
jastice: I was thinking this should be whatever the UI displays. Could be method name, fqn or some string describing a requirement, depending on how test framework organizes its tests. Ideas on the name?

jvican: We use `displayName` in `buildTarget`, so I propose we're consistent with that and rename it accordingly.

docs/bsp.md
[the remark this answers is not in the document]
jvican: > A plugin that runs some tasks to extract build structure, can run particular tasks and report success/failure.
  
It is a point of friction because you're not defining it in the build. In this scenario, you cannot even comply with the BSP connection protocol or other requirements in the protocol.
  
I'm inclined to not support this use case. A BSP server implementor should at least have access to the full universe of a build tool, it just cannot be implemented as a plugin.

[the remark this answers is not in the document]
jastice: > In this scenario, you cannot even comply with the BSP connection protocol or other requirements in the protocol.

Why not?

I think the discussion is sort of moot anyway. If a feature is not technically necessary for other features to work, implementors may or may not bother with it, whatever the spec says. I just think we better acknowledge this in the spec directly.

docs/bsp.md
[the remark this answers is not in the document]
jastice: Okay, let's make it a strict requirement. The notifications are still designed such that clients may ignore the start if they don't have a good way of reporting them.

docs/bsp.md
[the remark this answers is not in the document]
jastice: > Before this PR, sending CompileReport and TestReport was mandatory. I believe we should stick to the same semantics. I would prefer to discuss whether we should make this change or not in another PR, so that we don't block merging.

Sure. I made it optional to be in-line with task progress notifications in general, but agree to discuss in separate ticket.

docs/bsp.md
```diff
@@ -902,14 +1077,69 @@ trait TestReport {
...
-during compilation before completing the response.
+The server may send any number of `build/task*`, `build/publishDiagnostics` and `build/logMessage` notifications
+to communicate about tasks triggered by the request to the client.
+
+#### Test Unit Notifications
+
+The server may inform about individual test units in task notifications that reference the originating task in their `taskId`.
+For example, the server can send a `taskStart`/`taskFinish` for each test suite in a target, and likewise for each test in the suite.
+The server's implementation decides the granularity at which tests are reported. For example, if it only has information 
+about all the tests in a suite at a time, it could report a TestFinish for each test once the suite is done.
```
jvican: Can we update the scaladoc of `TestStart` and `TestFinish`?

jastice: What do you mean? :)

jvican:
It looks like CI is failing @jastice

Issue #42, opened 2018-10-25
Initialization request, additional information


jastice:
As discussed, we want additional data in the `InitializeBuildResult`:

* server info: String containing basic information about the server (typically name and version). This is primarily useful for debugging purposes
* bsp spec version: The version of the bsp specification this server implements (semver string)

Questions that I just came up with:
* should we have an arbitrary json data field that the server can fill as they please?
* should servers be able to communicate multiple bsp spec versions? This would require some further version negotiation in the protocol. Probably this is best left to server startup.

jvican:
> should we have an arbitrary json data field that the server can fill as they please?
  
In my opinion, this is always good.
  
> should servers be able to communicate multiple bsp spec versions? This would require some further version negotiation in the protocol. Probably this is best left to server startup.
  
Yes, I would leave it for server startup, and I would also leave the server info (name and version) there, adding only the BSP version to `build/initialize`. I'll expand on this point in the proposal to connect to the BSP server that I'm working on and will be submitted today.

olafurpg:
I agree with @jvican.

How about the following data structures?
```scala
trait InitializeBuildParams {
  def info: InitializeInfo
}
trait InitializeBuildResult {
  def info: InitializeInfo
}
trait InitializeInfo {
  /** BSP version that the client/server speaks. */
  def bspVersion: String
  /** Name of the client/server, can include the version number of the client application.
    * Can be displayed to the user to notify an an established connection. */
  def name: String
  /** Additional metadata about the client/server. */
  def data: Option[Json]
}
```

jastice:
@olafurpg why the nested `info`? Seems like it could go all directly into the params

olafurpg:
It was to share the same fields between the params and result types, otherwise we have to copy the `bspVersion,name,data` fields + docstrings in both objects. I don't feel strongly about it however.

jastice:
Makes sense, but I think request params and result data should be evolvable individually

jvican:
I agree; it's not clear to me it's safe to infer they will always be the same. I would just inline the fields.

olafurpg:
Sounds good to me, let's inline the fields 👍

olafurpg:
@jvican can we include this change in #53? It's related to the `BspConnectionDetails.`

jvican:
@olafurpg Yes! I'll add this information there.

PR #53, opened 2018-10-26
Add protocol to discover and connect to BSP servers

jvican:
A variant of https://github.com/scalacenter/bsp/issues/32 modified to support the following requirements:

1. Clients do not require beforehand knowledge about a specific server
   implementation to be able to connect to it.
2. Clients do not need to know the specifics of what kind of connections are
   supported by concrete build tools.
3. Clients can connect to build tools installed both at the machine and at the
   workspace level.
4. Multiple versions of the same build tool can have different connection
    mechanisms.
5. Multiple different BSP servers can run in the same workspace directory.
  
It now does substantially differ from the Jupyter way of addressing this problem.

If we agree on the wording of the spec I'll go ahead and implement the
code required for the JSON encoders/decoders of the BSP connection
details format.
  
Fixes https://github.com/scalacenter/bsp/issues/32


docs/bsp.md
```diff
@@ -1207,6 +1208,90 @@ Clients can use this information to reconstruct the tree of sbt meta builds. The
...
+The generated bsp file may contain machine-dependent information and must
+therefore not be checked into version control.
+
+
+### Connecting to BSP Servers
+
+BSP clients can connect to BSP servers by locating the JSON files in the
+standard BSP directory `<workspace-directory>/.bsp` in a project directory.
+
+When more than a BSP connection file is found, the client can use the
```
jastice: "more than one"

docs/bsp.md
```diff
@@ -1207,6 +1208,90 @@ Clients can use this information to reconstruct the tree of sbt meta builds. The
...
+and incorporates changes required to deal with  requirements above.
+
+### The BSP Connection Details
+
+The BSP connection details are modelled in a JSON object generated by a build
+tool command. The object contains the following fields to allow clients to
+connect to the server.
+
+```scala
+trait BspConnectionDetails {
```
jastice: We should have a field `protocolVersions: List[String]` to allow clients to decide if it can deal with the server even before connecting.

jastice: Should we cover the case of a resident server that we can connect to directly via tcp/ip rather than a command?

jastice: We need to specify some argument variable patterns that a client can pass to the server invocation for things like 
* socket file
* port
* bsp protocol version(?)
* flags (?)

jvican: Should we? I think it would be better to deal with this during the initialization handshake...

jvican: This is up to the build tool to implement. The build tool must generate be able to talk to a resident server (like gradle) or generate a script that the BSP client will run and will interact with the resident server.

docs/bsp.md
```diff
@@ -1207,6 +1208,92 @@ Clients can use this information to reconstruct the tree of sbt meta builds. The
...
+  /** A sequence of commands to connect to the BSP server that
+    * can be executed via system processes. */
+  def argv: List[String]
+}
+```
+
+Every build tool supporting BSP must implement a build-tool-specific command to
+generate the BSP connection details. The `argv` field in the connection details
+must be runnable via system process and can use whichever connection protocol
+is desired (UNIX Sockets, Windows Named Pipes, TCP, et cetera). The BSP server
```
olafurpg: Must the server and client communicate via stdin/stdout or can they use a custom connection protocol? The paragraph says both 

- "whichever connection protocol" and
- "will write messages to `stdout` and will receive messages via `stdin`"

docs/bsp.md
olafurpg: I think it's better to keep `Initialize{Params,Result}.bspVersion` and require a connection.

docs/bsp.md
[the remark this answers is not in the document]
olafurpg: @jastice beat me to it 😅

docs/bsp.md
jastice: True, that would work, though having it available at this point would make it easier to select/reject a connection spec without connecting.

jvican: Yes, this is a critical part, good point. I'm still rumbling how to support this in a principled way. I'll comment on it tonight.

docs/bsp.md
```diff
@@ -1207,6 +1208,90 @@ Clients can use this information to reconstruct the tree of sbt meta builds. The
...
+
+
+### Connecting to BSP Servers
+
+BSP clients can connect to BSP servers by locating the JSON files in the
+standard BSP directory `<workspace-directory>/.bsp` in a project directory.
+
+When more than a BSP connection file is found, the client can use the
+connection metadata to select only the BSP servers it's interested in. If there
+are still ambiguities, the BSP client is free to choose how to react.
```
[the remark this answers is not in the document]
olafurpg: edit:

> free to choose how to react, for example by asking the end user to select a build server.

jvican:
> I find it unsatisfying that we require users to generate .bsp/*.json files for every project directory
  
It's only done per workspace directory, not project directory.
  
> I believe the "when no BSP connection file is found" situation will be a frequent pain points for users.
  
This is why I've added that it's optional for the servers to generate the BSP details while they are running on a workspace. Once you run a build tool for something as simple as `bloop about`, the server would have already written the details to connect to it.

docs/bsp.md
```diff
@@ -1207,6 +1208,90 @@ Clients can use this information to reconstruct the tree of sbt meta builds. The
...
+```
+
+Every build tool supporting BSP must implement a build-tool-specific command to
+generate the BSP connection details.
+
+When such command is invoked, the build tool generates the connection details
+and writes them to a file located in a standard BSP directory
+`<workspace-directory>/.bsp`. The `.bsp/` directory must be at the root of the
+workspace base directory and the names of the files must be unique per build
+tool and version.
```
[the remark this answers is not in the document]
jvican: We can remove this requirement, it's not a big deal.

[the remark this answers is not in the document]
jastice: The wording jsut specifies that the name has to be unique per tool/version. How the tool achieves that is up to the tool. If there's one global installation it could just overwrite the old json, if there's parallel installations (eg development versions) it can create them with any arbitrary filename.

docs/bsp.md
```diff
@@ -1207,6 +1208,90 @@ Clients can use this information to reconstruct the tree of sbt meta builds. The
...
+ "languages": ["scala", "javascript", "rust"],
+ "argv": ["/usr/bin/python3", "my-build-tool", "bsp", "--protocol", "tcp", "--host", "localhost", "--port", "3141"]
+}
+```
+
+The generated bsp file may contain machine-dependent information and must
+therefore not be checked into version control.
+
+
+### Connecting to BSP Servers
```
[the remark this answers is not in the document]
jvican: I usually capitalize all nouns in the titles. Do you want to use lowercases instead?

docs/bsp.md
jvican: This is an error on my side, I wrote the wrong thing, I'll fix it.

jastice:
> > I believe the "when no BSP connection file is found" situation will be a frequent pain points for users.
> 
> This is why I've added that it's optional for the servers to generate the BSP details while they are running on a workspace. Once you run a build tool for something as simple as `bloop about`, the server would have already written the details to connect to it.

I believe the point is that we'd like to be able to import a project from bsp without the additional manual step of generating the config, and that a global config like with jupyter can take care of this. I think we should explore this in the future, but only once we get the basics as described in this proposal working.

olafurpg:
> It's only done per workspace directory, not project directory.

whoops, that's what I meant 😅 the `dev` directory in my machine contains 667 workspaces, that a lot of potentially unnecessary `.bsp/` directories if we supported global installation like Jupyter.

> This is why I've added that it's optional for the servers to generate the BSP details while they are running on a workspace.

If the norm becomes that clients hardcode build-specific connection commands then they might as well connect directly to the server without generating the bsp connection files.

olafurpg:
I agree this proposal is a good first step and global installation can be left for the future.

jvican:
> and that a global config like with jupyter can take care of this
  
IMO this is not possible because it goes against the requirement that the protocol must work for both tools installed globally or installed in a project. You cannot write static files of how to connect to a BSP server if your tool works at the project level. This is what has convinced me that having a global configuration directory, as Jupyter does, is not a good idea.

docs/bsp.md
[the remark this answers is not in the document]
olafurpg: There are two separate headers with almost identical titles

https://github.com/scalacenter/bsp/pull/53/files#diff-427028834811005be19d2a71da45dd28R1211
https://github.com/scalacenter/bsp/pull/53/files#diff-427028834811005be19d2a71da45dd28R1278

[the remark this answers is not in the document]
jvican: Oh, I missed this point. I'll update the titles.

jastice:
> IMO this is not possible because it goes against the requirement that the protocol must work for both tools installed globally or installed in a project

I don't see a fundamental reason why we can't have both. I will open a new ticket to discuss this so we don't sidetrack this PR.

docs/bsp.md
jvican: That’s exactly what I wanted to avoid, the connection file was in the original bsp ticket, but it looks like we’ll need it after all 😄

jvican: I've tentatively added it to the JSON object, but I'm happy to remove it if we find consensus.

docs/bsp.md
```diff
@@ -1288,6 +1288,10 @@ files:
 Build tools can decide where to write the connection files depending on the
 specific implementation details so long as the previous requirements are met.
 
+The `$XDG_DATA_DIRS` is defined in the [XDG Base Directory
+Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html).
+For Windows systems, the global location must be `%APPDATA%/.bsp/`.
```
jastice: why not also `%PROGRAMDATA%\jupyter\kernels`?

olafurpg: Can we maybe have a pretty table like kernelspec? 😇 

screen shot 2018-10-28 at 09 09 28

docs/bsp.md
```diff
@@ -1207,6 +1208,198 @@ Clients can use this information to reconstruct the tree of sbt meta builds. The
...
+ "argv": ["my-build-tool", "bsp"]
+}
+```
+
+#### Default Locations for BSP Connection Files
+
+There are two default locations where build tools can write BSP connection
+files:
+
+1. A global `.bsp/` directory located at a globally accessible location
```
olafurpg: nit: can we use `bsp` instead of `.bsp` for the global directory?

docs/bsp.md
```diff
@@ -1207,6 +1208,198 @@ Clients can use this information to reconstruct the tree of sbt meta builds. The
...
 simplify the data processing on the client side.
 
+## BSP Connection Protocol
+
+The Build Server Protocol defines a standard convention for clients to
+connect with BSP servers. This protocol has been designed such that:
+
+1. Clients do not require beforehand knowledge about a specific server
+   implementation to be able to connect to it.
+1. Clients do not need to know the specifics of what kind of connections are
```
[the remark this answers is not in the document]
jvican: The difference is not correctly explained in the items, so I'll rename it. I want to differentiate between being:
  
1. Connection-agnostic.
2. Build-tool-agnostic.
  
Jupyter is kernel-agnostic (our equivalent of build-tool-agnostic) but not connection-agnostic (it forces to use the zeromq connection details in the connection file). I want to emphasize here that this BSP connection protocol is both connection-agnostic *and* build-tool-agnostic, which is an improvement over the state of the art in Jupyter.
  
I'll reword these list items accordingly

docs/bsp.md
```diff
@@ -1207,6 +1208,198 @@ Clients can use this information to reconstruct the tree of sbt meta builds. The
...
+   supported by concrete build tools.
+1. Clients can connect to build tools installed both at the machine and at the
+   workspace level.
+1. Multiple versions of the same build tool can have different connection
```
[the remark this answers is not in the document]
jvican: Agreed, I'll remove it from this list but I'll add a paragraph making this observation explicit. I find it important that build tool implementors know this detail upfront.

docs/bsp.md
```diff
@@ -1207,6 +1208,198 @@ Clients can use this information to reconstruct the tree of sbt meta builds. The
...
+   mechanisms.
+1. Multiple different BSP servers can run in the same workspace directory.
+
+The following protocol to discover and connect to BSP servers is inspired by
+[Jupyter's
+Kernelspec](https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernelspecs),
+even though it has some notorious differences that are explained below.
```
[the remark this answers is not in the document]
jvican: I'm not sure we should remove the adjective, I want to stress that the differences are fundamental rather than tweaks. Do you want to propose another adjective instead?
  
Maybe the following formulation would be better?
  
> The following connection protocol is inspired by [Jupyter's Kernelspec](https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernelspecs) and improves on some of its aspects to abide by the protocol requirements stated above.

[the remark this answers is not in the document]
olafurpg: Your reformulation is fine. "notable" or "noteworthy" could work as well. "notorious differences" is not a positive thing

> famous or well known, typically for some bad quality or deed.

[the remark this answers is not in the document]
jvican: Oh thanks! I didn't know it had a negative connotation.

docs/bsp.md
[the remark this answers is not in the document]
jvican: Yes, that's true. I guess I'm considering zeromq as a further requirement in the connection protocol. stdin/stdout is the natural way of communicating when you shell out via system process, so you don't require your client to use an external protocol to connect to the process you just spawned.
  
I think this is an important detail that is not obvious from reading the protocol, if you think it's not I can conflate the two points in one.

olafurpg:
> Connection-agnostic.

I'm not convinced that "connection agnostic" is the best way to sell stdin/stdout. It's enough to say 

> The client spawns a build server process and communicates via standard input/output streams. Standard input/output streams work uniformly across different operating systems and are available on most runtimes without the use of external libraries.


----


> I'd be happy to go through the document and remove explicit references to Jupyter.

I think that's a good idea, the differences between kernelspec and BSP connections are large enough that the comparison doesn't help much.

jvican:
@olafurpg @jastice thanks for your feedback. I've addressed in the new revision, that rebases and adds two commits, one of them refining the specification and the other one adding the initialization information requested in #42

jastice:
Great work! Let's get this implemented!

docs/bsp.md
```diff
@@ -1308,6 +1313,23 @@ object ScalaPlatform {
 }
 ```
 
+#### Scala Test Params
+
+`ScalaTestParams` is a basic data structure that contains scala-specific
```
olafurpg: s/is a basic data structure that/

docs/bsp.md
```diff
@@ -1308,6 +1313,23 @@ object ScalaPlatform {
...
+metadata for testing Scala targets. This metadata is embedded in the `data:
+Option[Json]` field of the `buildTarget/test` request.
+
+```scala
+trait ScalaTestParams {
+  /** The test classes to be run in this test execution.
+    * It is the result of `buildTarget/scalaTestClasses`. */
+  def testClasses: List[ScalaTestClassesItem]
```
olafurpg: Does it make sense to have this be `Option[List[ScalaTestClassesItem]]`? To distinguish "run 0 test classes" and "run every test class"

docs/bsp.md
```diff
@@ -1308,6 +1313,23 @@ object ScalaPlatform {
...
+
+  /** Additional client-specific data that the server can parse and understand. */
+  def data: Option[Json]
```
olafurpg: Can we remove this field? If the client wants to send additional data then it can use `TestParams.data`, there is no need for double `.data.data` nesting.

olafurpg: Addressed in latest commit.

bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaExtension.xtend
```diff
@@ -27,7 +25,6 @@ class ScalaBuildTarget {
 @JsonRpcData
 class ScalaTestParams {
   @NonNull List<ScalaTestClassesItem> testClasses
-  @JsonAdapter(JsonElementTypeAdapter.Factory) Object data
```
olafurpg: Missing spec + bsp4s changes.

docs/bsp.md
jvican: I think it doesn't make sense. If the client is sending a test request, it has to test something, otherwise why do the request in the first place?

olafurpg: `TestParams.targets` includes the targets, I expect that to run all test classes defined in that target. Imagine that `ScalaTestParams.testOptions: List[String]` gets added down the road and I want to customize it, then I don't want to have to provide `testClasses: []`

PR #68, opened 2018-11-10
Rework default locations for connection files based on OS guidelines.

jastice:
Account for both user and system installations in base directory choices:

* for Unix user installations, use $XDG_DATA_HOME
* for Mac, use `~/Library/Application Support` and `Library/Application Support` in addition to unix paths
* for Windows user installations, user `%LOCALAPPDATA%`, which is a user data directory that is local to the machine, and `%PROGRAMDATA%` for system installs.


docs/bsp.md
```diff
@@ -1586,27 +1586,35 @@ This is an example of a BSP connection file:
...
-
-The user location has always higher priority than the system location, so if
-a client finds a BSP connection file that meets its criteria inside a user
-location it must pick it over other BSP connection files in the system
+1. `%LOCALAPPDATA%` and `%PROGRAMDATA%` are defined by the 
+[Windows Documentation](https://docs.microsoft.com/en-gb/windows/desktop/shell/csidl) 
+(see also: [Default Known Folders](https://docs.microsoft.com/en-gb/windows/desktop/shell/knownfolderid))
+1. on Macs, both standard macOS and Unix directories are supported
+
+The workspace location always has higher priority than the user or system location, so if
```
olafurpg: Unrelated to this PR since the paragraph was there before, but still want to comment that I'm on the fence about specifying this requirement here. Imagine I have a workspace with two different build tools (npm + sbt) and one build server is installed user-wide and other is project-specific. I think the client should be free to interpret the priority as it pleases.

jvican: I’ll act on this in an independent PR, I’m batching small changes I want to add to the protocol.

docs/bsp.md
```diff
@@ -1586,27 +1586,35 @@ This is an example of a BSP connection file:
...
-|--------|----------------------------------------|-------------------------------------------|
-| System | `$XDG_DATA_DIRS/bsp/`                  | `%PROGRAMDATA%/bsp/`                      |
-| User   | `<workspace-dir>/.bsp/`                | `<workspace-dir>/.bsp/`                   |
+A BSP connection file can be located in a number of locations. BSP connection files
+may be located in the project workspace, or for bsp servers installed locally, in a system-specific
+directory in the user's data system's data directories:
+
+|           | Unix + Mac                                         | Windows                    |
+|-----------|----------------------------------------------------|----------------------------|
+| Workspace | `<workspace-dir>/.bsp/`                            | `<workspace-dir>\.bsp`     |
```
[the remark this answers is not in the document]
jvican: Let's add the trailing `/` everywhere

PR #70, opened 2018-11-10
Add a simple mock BSP server and tests

jastice:
* enable end-to-end tests that don't depend on bloop
* a simple way for clients to test against a reference server implementation of current spec
* a library that allows to construct servers that emulate specific scenarios


build.sbt
```diff
@@ -95,4 +95,11 @@ lazy val tests = project
       "org.scalatest" %% "scalatest" % "3.0.5"
     )
   )
-  .dependsOn(bsp4s, bsp4j)
+  .dependsOn(bsp4s, bsp4j, mockServer)
+
+lazy val mockServer = project
```
olafurpg: Can this be moved into the `tests` module?

olafurpg: Or is the intention is to publish this module eventually?

jastice: The intention is to publish it to have a stand-alone testing app. That's my motivating use-case actually.

olafurpg: Sounds good, conventionally those kinds of modules are called "testkit". 

Out of curiosity, how come you decided to use bsp4s instead of bsp4j?

jastice: Not really a testkit at this point. But maybe it will be.
bsp4s just seemed a bit nicer and it's already there so why not?

jastice:
While using the mock server for testing, I found bsp4j does not deserialize enums correctly. I added the fix and some tests to this PR.

bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildTargetEventKind.java
```diff
@@ -1,5 +1,9 @@
 package ch.epfl.scala.bsp4j;
 
+import com.google.gson.annotations.JsonAdapter;
+import org.eclipse.lsp4j.jsonrpc.json.adapters.EnumTypeAdapter;
+
+@JsonAdapter(EnumTypeAdapter.Factory.class)
```
olafurpg: Good catch!

olafurpg:
The first commit starts with WIP, I'm personally not strict on commit message conventions but some people are so I leave it to y'all to merge this

PR #71, opened 2018-11-13
Add `reset` to `buildTarget/publishDiagnostics`

jvican:
The previous solution did not allow the client to handle the lifetime of
the diagnostics correctly. For example, when errors disappear in a
buffer, there is no way for the client to know that it should clean
them. Taking another approach in the lifetime handling brings us to
other problems like missing present warnings that are not present during
incremental compilation, see https://github.com/scalacenter/bloop/issues/696

After a discussion with Olafur, we have come up with the following
solution. Reset is a flag that allows servers to tell clients how they
should use the diagnostics. When `reset` is enabled, the behavior is the
same as in the Language Server Protocol. When it isn't enabled, it
allows the build tools to stream errors coming from, say, a compilation
run to the client and letting the client accumulate all of those until
the end of the compilation request.

The following commit makes changes to the code and the protocol spec to
reflect this new workflow.

Co-authored-by: Olafur Pall Geirsson <olafurpg@gmail.com>

### DIFF
--- a/docs/bsp.md
+++ b/docs/bsp.md
@@ -598,11 +598,12 @@ trait PublishDiagnosticsParams {
 
 where `Diagnostic` is defined as it is in LSP.
 
-When `reset` is enabled, the client must clean all previous diagnostics
+When `reset` is true, the client must clean all previous diagnostics
 associated with the same `textDocument` and `buildTarget` and set instead the
-diagnostics in the request. This is the same behaviour as `BuildTargetParams`
-in the LSP. When `reset` is disabled, the diagnostics are added to the last
-active diagnostics, allowing build tools to stream diagnostics to the client.
+diagnostics in the request. This is the same behaviour as
+`PublishDiagnosticsParams` in the LSP. When `reset` is false, the diagnostics
+are added to the last active diagnostics, allowing build tools to stream
+diagnostics to the client.
 
 It is the server's responsibility to manage the lifetime of the diagnostics by
 using the appropriate value in the `reset` field. Clients generate new
--- a/docs/bsp.md
+++ b/docs/bsp.md
@@ -568,10 +568,6 @@ defined it in the request that triggered this notification.
 The Diagnostics notification are sent from the server to the client to signal results of validation
 runs.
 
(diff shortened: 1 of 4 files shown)


jvican:
Note that this commit also turns `diagnostics` into a non-null list of diagnostics as per the BSP specification.

docs/bsp.md
```diff
@@ -593,14 +589,29 @@ trait PublishDiagnosticsParams {
...
 ```
 
 where `Diagnostic` is defined as it is in LSP.
 
-The definition of `PublishDiagnosticsParams` is similar to LSP's but contains the addition of an
-optional `originId` field. Clients can use this id to know which request originated the
-notification. This field will be defined if the client defined it in the original request that
+When `reset` is enabled, the client must clean all previous diagnostics
+associated with the same `textDocument` and `buildTarget` and set instead the
+diagnostics in the request. This is the same behaviour as `BuildTargetParams`
```
olafurpg: You mean `PublishDiagnosticsParams`?

docs/bsp.md
```diff
@@ -593,14 +589,29 @@ trait PublishDiagnosticsParams {
...
+  def reset: Boolean
 }
 ```
 
 where `Diagnostic` is defined as it is in LSP.
 
-The definition of `PublishDiagnosticsParams` is similar to LSP's but contains the addition of an
-optional `originId` field. Clients can use this id to know which request originated the
-notification. This field will be defined if the client defined it in the original request that
+When `reset` is enabled, the client must clean all previous diagnostics
```
olafurpg: s/enabled/true/

docs/bsp.md
```diff
@@ -593,14 +589,29 @@ trait PublishDiagnosticsParams {
...
+associated with the same `textDocument` and `buildTarget` and set instead the
+diagnostics in the request. This is the same behaviour as `BuildTargetParams`
+in the LSP. When `reset` is disabled, the diagnostics are added to the last
```
olafurpg: s/disabled/false/

jastice:
LGTM

PR #73, opened 2018-11-14
Improve explanation of `Task Id` section

jvican:
When implementing the task id bits in bsp v2, I looked at the
specification and found it confusing.

This commit hopefully improves the explanation of `Task Id` and stresses
that the `id` must be unique.

### DIFF
--- a/docs/bsp.md
+++ b/docs/bsp.md
@@ -251,12 +251,12 @@ trait BuildTargetIdentifer {
 
 ### Task Id 
 
-The Task Id allows clients to uniquely identify a resource and establish a client-parent
-relationship with another id.
+The Task Id allows clients to *uniquely* identify a BSP task and establish a client-parent
+relationship with another task id.
 
 ```scala
 trait TaskId {
-  /** The id */
+  /** The id, it has to be unique. */
   def id: String
 
   /** The parent id. */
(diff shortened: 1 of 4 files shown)


docs/bsp.md
```diff
@@ -251,24 +251,22 @@ trait BuildTargetIdentifer {
...
-The Task Id allows clients to uniquely identify a resource and establish a client-parent
-relationship with another id.
+The Task Id allows clients to *uniquely* identify a BSP task and establish a client-parent
+relationship with another task id.
 
 ```scala
 trait TaskId {
-  /** The id */
+  /** The id, it has to be unique. */
   def id: String
```
jvican: I've been tempted to make this a `URI` instead of a `String`, but I'm not sure it's the way to go. From the build tool side, it seems simpler to me to create a unique ID with a `URI` than creating an hexadecimal random string of a concrete length.
  
What do others think about this?

olafurpg: Proposed reformulation

> A unique identifier

docs/bsp.md
```diff
@@ -251,24 +251,22 @@ trait BuildTargetIdentifer {
...
   def parent: Option[String]
 }
 ```
 
-A task id can represent any child-parent relationship established by the build tool.
-
-An example of use of task ids is logs, where BSP clients can use the task ids of logs
-to improve their readability in the user interface. Clients can show logs in a tree fashion, for
-example, or with dropdowns.
+Clients can use Task Ids to improve the user experience. For example, a client can display
```
olafurpg: > to improve the user experience

This can be said about every feature in BSP, I would leave it out.

I propose to remove this paragraph entirely and update the `parent` docstring to say

> The parent task id, if any. A non-empty parent id means this task is a sub-task of the parent task. The child-parent relationship of tasks makes it possible to render tasks in a tree-like user interface.

docs/bsp.md
jastice: @jvican a random unique string is straightforward (eg random UUID or or even just random long), transforming it to a URI is an additional burden and not necessary

PR #74, opened 2018-11-14
Make `displayName` a required field

jvican:
As per BSP v1.0.


jastice:
Should it be required actually? displayName is only required for UX, not protocol functioning

jvican:
I wanted to have it compatible with bsp 1.0, but I guesss we can live with this being optional

PR #75, opened 2018-11-29
Add a logo

jastice:


jastice:
Our icon specialist made it :)
My idea was to have it in a top-level dir and later maybe include it with smaller icons in the libraries as a way for bsp implementors to use.

olafurpg:
I don't feel strongly about the directory, I think it's great to have a logo 👍 Thank you for arranging this @jastice

PR #80, opened 2019-01-16
Add buildTarget/Sources request to mock server and MockSuite

jastice:


testkit/src/main/scala/ch/epfl/scala/bsp/mock/HappyMockServer.scala
```diff
@@ -176,4 +201,6 @@ class HappyMockServer(base: File, val logger: Logger, implicit val client: Langu
   def uriInTarget(target: BuildTargetIdentifier, filePath: String): Uri =
     Uri(target1.uri.toPath.toUri.resolve(filePath))
 
+  private def asDirUri(path: Path): Uri =
+    Uri(path.toUri.toString + "/")
```
jastice: I find this really ugly and unintuitive :)

PR #82, opened 2019-01-22
Add dataKind field 

jastice:
* Add `dataKind` field to all objects containing a `data` field.
* Remove `classpath` in `SbtBuildTarget`
* Rename TestStarted / TestFinished classes in bsp4j to TestStart/TestFinish


jvican:
Why is the data kind field needed?

jastice:
> Why is the data kind field needed in the initialization request params/response?

In the case that a client specifies a specific format that the server may want to handle. I don't really have a strong opinion if a dataKind in this case is helpful, likewise for the `InitializeBuildResult`.

Issue #76, opened 2018-12-03
Spec gotcha: directory entry must end with `/`


jastice:
`SourceItem` specifies that a directory entry must end with `/`. This is a detail that is likely to be missed by implementors of the spec, both client and server and requires looking at the string to determine if something is a directory. I suggest making it an explicit field in `SourceItem` (and probably `DependencySourceItem` as well.

```
trait SourceItem {
  /** Either a text document or a directory. A directory entry must end with a forward
    * slash "/" and a directory entry implies that every nested text document within the 
    * directory belongs to this source item.
    */
  def uri: Uri

  /** Indicates if this source is automatically generated by the build and is not
    * intended to be manually edited by the user. */
  def generated: Boolean
}
```

olafurpg:
The "directory ends with slash" convention was borrowed from `nio.file.Path` but it's quite universal I think. I like it because by construction avoids stuff like

```
uri: "file:///Users/foo.scala"
directory: true
```

I don't feel strongly about this however.

jastice:
any opinion @jvican ?

jvican:
I like this convention and I think there's no benefit in splitting up the entities to represent files and directories. You make a good point that we should improve this invariant in the specification so that implementors and clients know about it. I would get the contents of that comment and put it in an independent paragraph.

jastice:
Okay with me. My concern is that ideally the spec wouldn't need to be referred to much while implementing. But this may not be a major issue here.

jastice:
Just ran into this actually. Java `File` and `Path` don't differentiate between directory and file, and serialize the path to the same string without trailing `/` either way. I suspect other file system libraries do the same. I think implementors are likely to run into this.

jastice:
even the bsp API has this issue:

```
> Uri("file:///here/or/there/").toPath
res0: java.nio.file.Path = /here/or/there
```

So the only way to check if an uri is intended as directory is to check the string of the uri. I think this is unelegant and error-prone

jvican:
Yes that’s expected, you need to add the trailing slash yourself, we do that in bloop

jastice:
You expect it, but I expect this to cost most people implementing this a few hours of debugging :(

olafurpg:
[...]

PR #83, opened 2019-01-22
Add optional `isDirectory` flag to `SourceItem`

jastice:
This reduces ambiguity and likely errors in the intended meaning of the uri, as the trailing slash convention is prone to be overlooked by implementors.

Addresses #76


docs/bsp.md
```diff
@@ -717,6 +717,10 @@ trait SourceItem {
     * directory belongs to this source item.
     */
   def uri: Uri
+  
+  /** Specifies whether uri denotes a directory or regular file. 
+    * If this is not set, and the uri does not point to an existing file, it should be treated as directory by default. */
+  def isDirectory: Option[Boolean]
```
olafurpg: Can we make this a `kind: Int` field with possible values `FILE` or `DIRECTORY`? A boolean `isX` option is often an enum screaming to get out!

jastice: I'm not sure that makes sense in this case. Do we have anything else these paths could potentially be?

olafurpg: I think `if (!item.isDirectory)` is less clear than `if (item.kind == FILE)`. Also, with kind we could drop the `generated: Boolean` option and instead use 

```
enum Kind { FILE, DIRECTORY, GENERATED_FILE, GENERATED_DIRECTORY }
```

jastice: Agree on clarity, though boolean encoding is simpler.

generated and directory are orthogonal, so I don't think it makes sense to combine them into a single enum (though that is how it's done within IntelliJ)

Overall I'm not very emotional on this, so let's see what @jvican says

jvican: I prefer the enum encoding over the boolean to keep the API open and avoid API gotchas in the future. Otherwise, what `isDirectory` meant in the past could be broken with the introduction of new types in the future.

jastice: Okay, there's several ways we can do it:
* a field `kind` which is one of `FILE` or `DIRECTORY`, and keep `generated` Boolean.
* a field `kind` which is one of `FILE`, `DIRECTORY`, `GENERATED_FILE`, `GENERATED_DIRECTORY` and replaces `generated`. This has the drawback that if we add more kinds, expecially orthogonal kinds, it would bloat the possible enum values.
* separate field for `kind` and `generated`, both containing enums.
* a field `kinds` OR `tags` which can contain 0 or more of `FILE`, `DIRECTORY`, `GENERATED`, which replaces both booleans. Files would default to non-generated directories.

The tag approach is most flexible, but makes the API less explicit in declaring that a file/dir needs to be explicitly declared.

jvican: I like Olafur's suggestion to define it like `enum Kind { FILE, DIRECTORY, GENERATED_FILE, GENERATED_DIRECTORY }`

jastice: Why do you prefer it? It doesn't expand well when adding more attributes. Anyway, I've pushed a version using the `kinds` list for your consideration.

docs/bsp.md
```diff
@@ -726,8 +733,19 @@ trait SourceItem {
...
   def generated: Boolean
 }
+
+object SourceItemKind {
+  /** The source item references a normal file.  */
+  val File: Int = 1
+  /** The source item references a directory. */
+  val Directory: Int = 2
+  /** The source item is a generated file or directory. */
+  val Generated: Int = 3
```
jvican: Can we have `GeneratedFile` and `GeneratedDir` (and, in bsp4j, `GENERATED_FILE` and `GENERATED_DIR`)?

docs/bsp.md
```diff
@@ -706,18 +706,25 @@ Response:
...
   def uri: Uri
   
+  /** A kind denotes properties of the source item, such as whether it is file or directory, and generated.
+   * If no kinds are specified, the item referenced by `uri` should be assumed to be a regular directory. 
+   */
+  def kinds: List[SourceItemKind]  
+  
   /** Specifies whether uri denotes a directory or regular file. 
     * If this is not set, and the uri does not point to an existing file, it should be treated as directory by default. */
   def isDirectory: Option[Boolean]
```
jvican: Should we remove this approach now?

docs/bsp.md
olafurpg: `List[Kind]` allows representing invalid values like `[File,Directory,File]`. I like `enum Kind { FILE, DIRECTORY, GENERATED_FILE, GENERATED_DIRECTORY }` since it represents only valid values. I am OK with doubling the number of enum values for every new attribute, I don't expect we'll add too many attributes in the future.

docs/bsp.md
jastice: I'd say either we have a simple value field with the items 'File`, `Directory`, `GeneratedFile`, `GeneratedDirectory`, or a list of "tags" like this with onle 'File`, `Directory`, `Generated`. I prefer the list because being generated is orthogonal to being a file or directory.

docs/bsp.md
jastice: The isDirectory? Yes, I forgot to remove it.

jastice:
> I like better this explicit API. I've suggested some more changes. We should also mention in the docs that we don't represent symbolic/hard links at the API level (instead, they are a type of a file) and that the client must handle them adequately depending on the OS/file system it runs on.

Agree.

> One thing we must discuss with @olafurpg too is if it makes sense to have a `GENERATED_DIRECTORY` at all. For example, one could say that only files can be generated and directories shouldn't. However, I think it still makes sense to have `GENERATED_DIRECTORY` because BSP clients would usually like to warn their users whenever they are creating a file inside a generated directory.

At least on IntelliJ side, there is only directories. A "generated directory" is assumed to contain only generated sources.

docs/bsp.md
jastice: Do you have any opinion @olafurpg

docs/bsp.md
jastice: Okay, let's go with that then.

docs/bsp.md
olafurpg: Sorry for the slow response! I agree the "generated" attribute is orthogonal to the item being a file or directory. I still prefer enumerating all possible valid values (file, dir, generated file, generated dir) instead of allowing invalid representations like `[File, Directory]`. One alternative is to keep `kind: FILE | DIRECTORY` and `generated: Boolean`. I don't have a strong opinion either way.

jvican: Thinking more about it, I'd favor the `kind: FILE | DIRECTORY` and `generated: Boolean`. The users of this API would more likely be interested in picking the type of the source more than in that being generated or not, so with this approach they only need to compare against `FILE` or `DIRECTORY` instead of doing two comparisons (e.g. `FILE && GENERATED_FILE`)

jastice: Agree, I went with this and pushed. Please have another quick look.

PR #84, opened 2019-01-29
Generator tests

jastice:
Includes previous PRs #82  #83 

I created ScalaCheck generators for all the bsp4j data classes and roundtrip serialization tests bsp4j -> bsp4s -> bsp4j

While writing the tests, I found a few inconsistencies between the APIs and spec. These are fixed in this PR.


docs/bsp.md
```diff
@@ -717,6 +736,10 @@ trait SourceItem {
     * directory belongs to this source item.
     */
   def uri: Uri
+  
+  /** Specifies whether uri denotes a directory or regular file. 
+    * If this is not set, and the uri does not point to an existing file, it should be treated as directory by default. */
+  def isDirectory: Option[Boolean]
```
olafurpg: Are we sure we want to use `isDirectory` instead of `kind`?

jastice: not feeling strongly, but as stated in #83 I think this is a case where we needn't expect expansion of types of files on this dimension :)

jvican:
I'll review this as soon as the other PRs are acted on and merged, to simplify the review of this PR :smile:

jastice:
I rebased onto the other PRs. Please have a look, but don't merge yet, I'm still updating as I write tests on IntelliJ side.

jastice:
Okay I consider it ready for merge after review. Please have another look @olafurpg @jvican

jastice:
> Note that this PR has several binary breaking changes. It's not a problem since we're in milestone phase but once we commit to a stable v2 we should avoid breaking changes.

absolutely. Part of the point of this was to suss out problems before setting on a stable v2. Which ideally would happen next week to make the IntelliJ bsp support public.

olafurpg:
I agree, it's fantastic to iron out so many issues before the stable release 👍

PR #87, opened 2019-03-13
Remove streaming of diagnostics

jvican:
In previous versions of BSP, we added a field `reset: Boolean` to
diagnostics that would allow the server to stream diagnostics to the
client, instead of sending them all at once.

The motivation for this feature is that the end user would get more
instant feedback than with the batched approach. The way this approach
worked is that the first diagnostic notification for a text document
identifier would have `reset: Boolean = true` and the rest would be set
to `false` preventing the client from clearing all diagnostics in the
same text document.

This commit instead removes this approach. This removal is motivated by
the real-world experience of using this diagnostics publishing strategy
in tools in the Scala ecosystem. There are two main reasons for its
removal:

1. Notifications in JSON-RPC have no ordering so they can arrive to the
client in whatever order they are sent/processed by the OS/client.
Therefore, the approach above could be not setting certain diagnostics
if they are sent before the diagnostic that sets `reset = true`. This
has happened to me several times while using these tools.

2. There are other ways we can provide instant feedback to the client.
Instead of streaming, the server can send small batches of diagnostics
for text document identifiers as the compiler goes through the code
(e.g. every time the compiler finishes a compilation phase).  This comes
at the cost of delivering more notifications to the client and minimizes
issues with race conditions when setting diagnostics.

Therefore, we stay with the LSP approach of sending diagnostics. It's
clearer and simpler for BSP clients.

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
@@ -235,13 +235,11 @@ class PublishDiagnosticsParams {
   @NonNull TextDocumentIdentifier textDocument
   @NonNull BuildTargetIdentifier buildTarget
   @NonNull List<Diagnostic> diagnostics
-  @NonNull Boolean reset
   String originId
-  new(@NonNull TextDocumentIdentifier textDocument, @NonNull BuildTargetIdentifier buildTarget, @NonNull List<Diagnostic> diagnostics, @NonNull Boolean reset) {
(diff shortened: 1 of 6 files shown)


propensive:
Ignore my initial thumbs-down (in case you saw it). The streaming approach was working well enough with Fury, but I think that your alternative approach would work just as well, and maybe better.

I don't fully understand point 1, but I think it corresponds to some nonfatal errors (just printed on stdout) I've been seeing during builds. Having the diagnostics come in progress updates would be better, as we could then disambiguate between parsing, typer and later errors.

jastice:
Shall we merge this then?

jvican:
Let's not merge, I have doubts about this approach and I now think streaming in diagnostics is inevitable. Will try to write down my thoughts on this soon

Issue #88, opened 2019-04-24
Allow running specific main class via the buildTarget/run


marek1840:
Currently, if the project contains more than one main class - there is no way to specify which should be run when the `buildTarget/run` is requested. 
It may be useful - probably mostly during the development process - to be able to create small ad hoc main class for whatever reason. Also, some application may want to specify various entry points with different functionality exposed not as a set of flags to be passed but by separate main classes.

As suggested here: https://github.com/scalacenter/bloop/issues/859#issuecomment-485875217

> we should change buildTarget/run to have two more fields:
> 
>     dataKind: String
>     data: Option[Json]
> 
> like `buildTarget/compile` and `buildTarget/test` have.
> 
> Like in `buildTarget/test`, we can define a Scala Main Classes request that is equivalent to the Scala Test classes request and pass that to the data field of the run request.

jvican:
Fixed by https://github.com/scalacenter/bsp/pull/89

PR #89, opened 2019-04-24
include optional 'data' field in buildTarget/run request

marek1840:
This allows the client to customize the application execution request (e.g. by specifying the main class as a desired entry point)

Addresses #88


jastice:
LGTM, thanks!

PR #94, opened 2019-07-25
Add DAP capabilities to BSP

marek1840:
This allows the client to run/debug code using the Debug Adapter Protocol

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/LaunchParametersDataKind.java
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/LaunchParametersDataKind.java
@@ -1,0 +1,6 @@
+package ch.epfl.scala.bsp4j;
+
+public final class LaunchParametersDataKind {
+    public static String SCALA_MAIN_CLASS = "scala-main-class";
+    public static String SCALA_TEST_SUITES = "scala-test-suites";
+}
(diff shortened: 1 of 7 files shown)


bsp4j/src/main/java/ch/epfl/scala/bsp4j/LaunchParametersDataKind.java
```diff
@@ -0,0 +1,5 @@
+package ch.epfl.scala.bsp4j;
+
+public final class LaunchParametersDataKind {
+    public static String scalaMainClass = "scala-main-class";
```
marek1840:
```suggestion
    public static String scalaMainClass = "scala-main-class";
    public static String scalaTestSuite = "scala-test-suite";
```

bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
```diff
@@ -622,12 +622,23 @@ object ScalaPlatform {
...
-@JsonCodec final case class BspConnectionDetails(
-  name: String,
-  argv: List[String],
-  version: String,
-  bspVersion: String,
-  languages: List[String]
-)
\ No newline at end of file
+object LaunchParametersDataKind {
+  val scalaMainClass = "scala-main-class"
```
marek1840:
```suggestion
  val ScalaMainClass = "scala-main-class"
  val ScalaTestSuite = "scala-test-suite"
```

marek1840:
This was superseded by #96 #97 and #98.
Those changes are much simpler as #96 and #97 are in no way related to the `DAP` and would probably be proposed somewhere in the future as they are natural extensions to the `BSP`.

#98 on the other hand introduces a simple way of communication that the process requested by the client using `run/test` endpoints is ready to accept connections and communicate in whatever way is appropriate for this debuggee (e.g. Java Debug Wire Protocol for jvm applications)

PR #96, opened 2019-09-06
add jvm options to ScalaTestParams

mzarnowski:
This allows the user to customize the JVM on which the tests are run. Useful when one wants to, for example, attach to this VM to debug the tests.

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaExtension.xtend
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaExtension.xtend
@@ -24,7 +24,17 @@ class ScalaBuildTarget {
 
 @JsonRpcData
 class ScalaTestParams {
-  List<ScalaTestClassesItem> testClasses
+  @NonNull List<ScalaTestClassesItem> testClasses
+  @NonNull List<String> jvmOptions
+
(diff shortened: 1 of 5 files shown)


jastice:
I think this is effectively superceded by #107. Should we close it?

marek1840:
> I think this is effectively superceded by #107. Should we close it?

This add the possibility of setting the jvmOptions (just like we do for the `Run` endpoint). 
The #107 in its current state allows one to fetch the jvmOptions so those seem as opposite features.

jastice:
> > I think this is effectively superceded by #107. Should we close it?
> 
> This add the possibility of running tests with specified jvmOptions (just like we do for the `Run` endpoint).
> The #107 in its current state allows one to fetch the jvmOptions so those seem as opposite features.

Oh right, okay, I see

bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaExtension.xtend
```diff
@@ -24,7 +24,17 @@ class ScalaBuildTarget {
 
 @JsonRpcData
 class ScalaTestParams {
-  List<ScalaTestClassesItem> testClasses
+  @NonNull List<ScalaTestClassesItem> testClasses
+  @NonNull List<String> jvmOptions
```
jastice: I think this field should be optional

jastice:
This can be closed as obsoleted by #144, right?

PR #99, opened 2019-09-11
Add DAP capabilities to BSP

marek1840:
This allows the client to run/debug code using the Debug Adapter Protocol

### DIFF
--- a/bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
+++ b/bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
@@ -624,12 +624,23 @@ object ScalaPlatform {
     parent: Option[BuildTargetIdentifier],
     children: List[BuildTargetIdentifier],
 )
-
-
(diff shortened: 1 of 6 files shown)


bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
```diff
@@ -622,12 +622,23 @@ object ScalaPlatform {
...
-@JsonCodec final case class BspConnectionDetails(
-  name: String,
-  argv: List[String],
-  version: String,
-  bspVersion: String,
-  languages: List[String]
-)
\ No newline at end of file
+object DebugSessionParamsDataKind {
+  val scalaMainClass = "scala-main-class"
```
jvican: Let's capitalise the names of these symbols, like we do in `BuildTargetDataKind`:
```scala

object BuildTargetDataKind {
  val Scala = "scala"
  val Sbt = "sbt"
}
```

PR #107, opened 2020-02-04
Add ExecutionEnvironment endpoint

tpasternak:
The new endpoint is useful when IDE needs to run unit tests using it's native
mechanism. We need to get classpath, workingDir and environment
variables from BSP server, if we actually treat it as the main
source of truth.


tpasternak:
If this PR is merged, the endpoint will be used to fix this issue in IntelliJ https://youtrack.jetbrains.com/issue/SCL-16934

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
```diff
@@ -530,6 +531,36 @@ class RunResult {
...
+  new(@NonNull BuildTargetIdentifier target, @NonNull List<String> classpath) {
+    this.target = target
+    this.classpath = classpath
+  }
+}
+
+@JsonRpcData
+class ExecutionEnvironmentResult {
+  @NonNull String workingDirectory
+  @NonNull Map<String, String> environmentVariables
```
olafurpg: Can we include Java options here? `List<String> javaOptions`, which include system properties like `-D...` and memory options like `-X...`

bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildServer.java
```diff
@@ -46,6 +46,9 @@
     @JsonRequest("buildTarget/cleanCache")
     CompletableFuture<CleanCacheResult> buildTargetCleanCache(CleanCacheParams params);
 
+    @JsonRequest("buildTarget/executionEnvironment")
+    CompletableFuture<ExecutionEnvironmentResult> getClasspath(ExecutionEnvironmentParams params);
```
olafurpg: Should we move this to `ScalaBuildServer` and call it `buildTarget/javaExecutionEnvironment`? We try to keep the language agnostic APIs out of the core BSP spec.

bsp4s/src/main/scala/ch/epfl/scala/bsp/endpoints/Endpoints.scala
```diff
@@ -45,6 +45,8 @@ trait BuildTarget {
   object scalaMainClasses
       extends Endpoint[ScalaMainClassesParams, ScalaMainClassesResult](
         "buildTarget/scalaMainClasses")
+  object scalaClasspath extends Endpoint[ExecutionEnvironmentParams, ExecutionEnvironmentResult]("buildTarget/executionEnvironment")
```
olafurpg:
```suggestion
  object executionEnvironment extends Endpoint[ExecutionEnvironmentParams, ExecutionEnvironmentResult]("buildTarget/executionEnvironment")
```
Or `javaExecutionEnvironment`

bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/ClasspathItem.java
```diff
@@ -0,0 +1,82 @@
+package ch.epfl.scala.bsp4j;
+
+import ch.epfl.scala.bsp4j.BuildTargetIdentifier;
+import java.util.List;
+import org.eclipse.lsp4j.jsonrpc.validation.NonNull;
+import org.eclipse.xtext.xbase.lib.Pure;
+import org.eclipse.xtext.xbase.lib.util.ToStringBuilder;
+
+@SuppressWarnings("all")
+public class ClasspathItem {
```
jvican: Should we call this `ClasspathEntry`? I find it more JVM idiomatic than `ClasspathItem` despite the convention of using `Item` across the build server protocol spec. I think this case merits an exception.

bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildServer.java
jvican: To add to that, what do you think about adding more specific execution environments instead of a single request? I don't think a build target has a unique execution environment. For example, the environment to run an application or run tests can be different. I would go for `buildTarget/javaRunEnvironment`  and `buildTarget/javaTestEnvironment` instead.

bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
```diff
@@ -576,6 +576,23 @@ object TestParamsDataKind {
...
 )
 
+
+@JsonCodec final case class ExecutionEnvironmentParams(
+    targets: List[BuildTargetIdentifier],
+    originId: Option[String]
+)
+
+@JsonCodec final case class ClasspathItem(
+    target: BuildTargetIdentifier,
```
jvican: Is the purpose of this abstraction to represent a classpath per build target? If so, the name should make that clearer, `ClasspathItem` suggests a `List[String]`.
  
In addition to that, the classpath is leaking some key information about build targets which could be considered internal. What happens if this classpath goes out-of-date from one compilation to another one? If we don't handle this use case, are we not assuming build tools can always generate a unique classpath that doesn't change? This assumption might be wrong if, for example, a build tool generates immutable jars after every compilation.

bsp4s/src/main/scala/ch/epfl/scala/bsp/endpoints/Endpoints.scala
jvican: Maybe `jvmExecutionEnvironment` would be more appropriate? The execution environment for JVM languages should be the same.

bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/ClasspathItem.java
tpasternak: I already renamed it to `TargetJavaExecutionEnvironmentItem`, because it contains more than just classpath (jvm options). Of course I can do the rename again. Would `JvmExecutionEnvironmentEntry` be OK in your opinion?

tpasternak: Oh, maybe `JvmEnvironmentEntry` or `JvmTestEnvironmentEntry` would better correspond to your other comments in this PR.

bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
tpasternak: > Is the purpose of this abstraction to represent a classpath per build target? If so, the name should make that clearer, ClasspathItem suggests a List[String].
Yes, that was right before I added jvmOptions to the class and renamed it. Now I think a dedicated case class for this is needed (`JvmEnvironmentEntry`)

> In addition to that, the classpath is leaking some key information about build targets which could be considered internal. What happens if this classpath goes out-of-date from one compilation to another one? If we don't handle this use case, are we not assuming build tools can always generate a unique classpath that doesn't change after compilation runs? This assumption might be wrong if, for example, a build tool generates immutable jars after every compilation.
My assumption is that, the IDE is doing this call right before the execution, but after compilation is finished, so it's up to the server to keep it up-to-date. Please correct me if I'm wrong.

tpasternak:
@jvican @olafurpg almost all names was changed, could you look at the whole changeset once more?

bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildServer.java
tpasternak: As for now, I made  `buildTarget/javaTestEnvironment` only

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
tpasternak: Done

bsp4s/src/main/scala/ch/epfl/scala/bsp/endpoints/Endpoints.scala
tpasternak: Changed it to jvmTestEnvironment, according to the previous comment

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
olafurpg: The current encoding doesn't allow unique working directories or environment variables per targets. Would it make sense to move those into `JvmEnvironmentEntry`? At the risk of shaving yaks, I think the name should end with "Item" instead of "Entry" and organized like this

```scala
case class JvmEnvironmentItem(workingDirectory: Path, environmentVariables: Map[String, String], classpath, ...)
class JvmEnvironmentResult(items: List[JvmEnvironmentItem])
```

I agree we should use "entry" for single elements of a classpath.

bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaBuildServer.java
```diff
@@ -14,4 +14,7 @@
 
     @JsonRequest("buildTarget/scalaMainClasses")
     CompletableFuture<ScalaMainClassesResult> buildTargetScalaMainClasses(ScalaMainClassesParams params);
+
+    @JsonRequest("buildTarget/jvmTestEnvironment")
+    CompletableFuture<JvmEnvironmentResult> jvmTestEnvironment(JvmEnvironmentParams params);
```
olafurpg: 👍 I like the new name, I don't have a strong opinion on `jvmTestEnvironment` vs. `jvmEnvironment`

bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
olafurpg: > What happens if this classpath goes out-of-date from one compilation to another one? 

I think you can treat it the same way `buildTargets/scalacOptions` already leaks internal information. IntelliJ and Metals are already using those values to assemble a classpaths. I believe the reason it's working currently for IntelliJ is that it uses a custom stable classes directory, which Bloop does not remove after the BSP session closes.

I don't know what the best solution is for manifest jars however. Maybe it's OK to create temporary files and leave them there forever 🤔 Maybe we should include in the `jvmTestEnvironment` metadata about whether the client should delete the classpath entries?

jvican: > My assumption is that, the IDE is doing this call right before the execution, but after compilation is finished, so it's up to the server to keep it up-to-date.

That sounds good, so long as we make this clear in the spec that the client should not cache this at any cost, then my concern is resolved. BTW, what do you think about updating the spec with these changes?

> I think you can treat it the same way buildTargets/scalacOptions already leaks internal information

Yup, my main concern is about caching and how the client should handle that, especially in the presence of manifest jars that are immutable (e.g. bazel). However, if this call is done right after every compilation, this issue should be resolved. The problem comes when the jars in the classpath needs to survive several compilation runs since they assume mutability.

bsp4s/src/main/scala/ch/epfl/scala/bsp/endpoints/Endpoints.scala
```diff
@@ -45,6 +45,8 @@ trait BuildTarget {
   object scalaMainClasses
       extends Endpoint[ScalaMainClassesParams, ScalaMainClassesResult](
         "buildTarget/scalaMainClasses")
+  object jvmTestEnvironment extends Endpoint[JvmEnvironmentParams, JvmEnvironmentResult]("buildTarget/jvmTestEnvironment")
```
jvican: Should we not have one for running too? e.g. `jvmRunEnvironment`?

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
tpasternak: This comes from my understanding of how it is done in Bloop. Working directory and environment variables are stored in State class, so if your endpoint looks like this:
```
    ifInitialized(None) { (state: State, logger: BspServerLogger) =>
      mapToProjects(params.targets, state) match {
```
the working directory is stored in `state.commonOptions.env` and environment variables are in `state.commonOptions.workingDirectory`. 

I don't know if it makes sense to put it into `Entry` class,  but it could. If you think it'll be better,  in current bloop implementation I'd just copy the same value to all entries in current bloop implementation. Pleas just let me know what's your view on this.

bsp4s/src/main/scala/ch/epfl/scala/bsp/endpoints/Endpoints.scala
olafurpg: Can you elaborate on the purpose of that? For the concrete use-case we have at the moment in IntelliJ, we don't have the need for separate run/test environment.

olafurpg: I would be fine with `jvmEnvironment`, which can generically handle any kind of `java -classpath ...` invocation

jvican: It's fine by me if you don't include `jvmRunEnvironment`, but `jvmTestEnvironment` should not be renamed to `jvmEnvironment` because the environment in `run` or `test` could be different

bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
tpasternak: > BTW, what do you think about updating the spec with these changes?

Of course, just doing it

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
olafurpg: In Bloop, each `Project` can have a unique working directory https://github.com/scalacenter/bloop/blob/dfd03f5333c2fc292db78bfcd4205a0cdd828967/frontend/src/main/scala/bloop/data/Project.scala#L66

It's the same situation in sbt, where `baseDirectory.in(run).value` is used for running in forked JVMs.

As for environment variables, Bloop currently only uses the environment variables that are passed from the terminal interface. However, for consistency with other configuration I don't see a reason why we should not allow per-target variables.

jvican: I agree with @olafurpg's suggestion that working directory and env variables could make more sense as part of the `JvmEnvironmentEntry` since they can be project-specific.

Regarding your comments of how we do it in bloop:
* We don't use `state.commonOptions.workingDirectory` as the working directory of the project we run but the `directory` configuration field that is configured at the project lefvel. 
* You're right we currently allow the client to pass the environment variables from the place the CLI is run via `state.commonOptions.env`, but in the future we will enable build tools to export environment variables to the configuration file on a per-project basis. This is a common idiom so we want to support it.

bsp4s/src/main/scala/ch/epfl/scala/bsp/endpoints/Endpoints.scala
olafurpg: fair enough, it can indeed be different in sbt :+1:

jvican:
Feel free to merge after the comments are addressed :+1:

olafurpg:
> BTW, now that we have defined the inputs for the run/test tasks, could you follow up to this PR and add the working directory, jvm options et al to ScalaRunParams and ScalaTestParams?

@jvican I propose we leave those to a separate PR since they are independent features. We should not burden @tpasternak with updating those in this PR.

jvican:
Yup, we're on the same page, I said "follow up to this PR" to mean "can you open a follow-up PR"

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
tpasternak: Ok, thanks, I'm just changing this

tpasternak:
There's still "Items" vs "Entries" issue. Please let me know what's your opinion for this

olafurpg:
I personally prefer "items" over "entries" but it's not a hill I will die on :)

tpasternak:
@jvican you suggested that "Entry" is more idiomatic to JVM, but it was a `ClasspathEntry`. Now it's extended and renamed to `JvmEnvironmentEntry`, so I guess the previous argument is not the case any more. Should I switch it back to `Item` to satisfy the BSP convention?

jvican:
@tpasternak Yes, sounds good to me 👍 I was suggesting to use entries only for the classpath entries (the independent elements that make a classpath). I agree we should use `Item` to follow the BSP convention

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
```diff
@@ -530,6 +531,39 @@ class RunResult {
   }
 }
 
+@JsonRpcData
```
jastice: These should be in a separate JvmExtension file, since JVM is language/environment-specific, and wouldn't need to be supported by all build servers

bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaBuildServer.java
```diff
@@ -14,4 +14,7 @@
 
     @JsonRequest("buildTarget/scalaMainClasses")
     CompletableFuture<ScalaMainClassesResult> buildTargetScalaMainClasses(ScalaMainClassesParams params);
+
+    @JsonRequest("buildTarget/jvmTestEnvironment")
```
jastice: Should be in a JVMBuildServer interface (which Scala one can possibly extend)

bsp4s/src/main/scala/ch/epfl/scala/bsp/endpoints/Endpoints.scala
jastice: I think if we have `jvmTestEnvironment` we should have a corresponding `jvmRunEnvironment` or just general `jvmEnvironment`, even if it's not needed for the motivating use case.

docs/specification.md
```diff
@@ -1308,3 +1308,44 @@ export interface CleanCacheResult {
...
 ```
+
+### Test Environment Request
+
+The JVM test environment request is sent from the client to the server in order to
+gather information required to launch a Java process. This is useful when the
+client wants to control the Java process execution, for example to enable custom
+Java agents or launch a custom main class during unit testing or debugging
+
+The data provided by this endpoint may change beteween compilations, so it should
```
jastice: typo: beteween

docs/specification.md
```diff
@@ -1308,3 +1308,44 @@ export interface CleanCacheResult {
...
+not be cached in any form. The client should ask for it right before test exection,
```
jastice: typo: exection

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
tpasternak: Done

bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaBuildServer.java
tpasternak: Done

docs/specification.md
tpasternak: Done, moved to jvm.md

docs/specification.md
tpasternak: Done, moved to jvm.md

tpasternak:
Sorry, I forgot to change the params/result type name when I was changing jvmEnvironment => jvmTestEnvironment. Just did it in  8cfc7d8

tpasternak:
I think all issues are covered now

jastice:
Thank you!

If I understand it right, if we add a general `jvmEnvironment` endpoint, it would cover the use case of #96 

I'd also like basic test coverage to ensure serialization between implementations is compatible.

Also note I've just merged #106 which may require a rebase

tpasternak:
> If I understand it right, if we add a general `jvmEnvironment` endpoint, it would cover the use case of #96
I don't think so, the endpoint in this PR is only to gather some data, but in #107 it's possible to set it when tests are executed via BSP

> I'd also like basic test coverage to ensure serialization between implementations is compatible.
Added here - is it enough? https://github.com/build-server-protocol/build-server-protocol/pull/107/commits/d3b67e20ab0a9726fa7b48f58fba590401260897

jastice:
> > If I understand it right, if we add a general `jvmEnvironment` endpoint, it would cover the use case of #96
> 
> I don't think so, the endpoint in this PR is only to gather some data, but in #107 it's possible to set it when tests are executed via BSP

Okay I see.

> Added here - is it enough? [d3b67e2](https://github.com/build-server-protocol/build-server-protocol/commit/d3b67e20ab0a9726fa7b48f58fba590401260897)

Great, that will do for now! We can add generators and serialization tests later.

docs/extensions/jvm.md
```diff
@@ -0,0 +1,50 @@
...
+id: jvm
+title: JVM Extension
+sidebar_label: JVM
+---
+
+The following section contains JVM-specific extensions to the build server
+protocol.
+
+
+## Test Environment Request
```
jastice: Any opinion if we should also add generic `buildTarget/jvmEnvironment` here as well?
@jvican @olafurpg

tpasternak:
> > Added here - is it enough? [d3b67e2](https://github.com/build-server-protocol/build-server-protocol/commit/d3b67e20ab0a9726fa7b48f58fba590401260897)
> 
> Great, that will do for now! We can add generators and serialization tests later.

Oh, I see. I haven't noticed there are already such tests. I can add it to this PR or open next one, please just let me know which you prefer.

jastice:
If you can add them now it's easiest :)

tpasternak:
Hi @jastice @jvican, what's your opinion now - is it ready to merge?

olafurpg:
This feature should now be available in the latest 2.0.0-M5 release

PR #108, opened 2020-02-24
 Include JvmTestEnvironment in Build Server Capabilities

tpasternak:
It was requested by @jastice here https://github.com/JetBrains/intellij-scala/pull/500#pullrequestreview-360335173


jastice:
Looks good, but a thing to consider is if language/env extensions shoudl declare their capabilities deparately from `BuildServerCapabilities`. Any opinion @olafurpg @jvican ?

olafurpg:
I don’t think that’s necessary. In Metals we declare lsp extensions in the initialize handshake.

tpasternak:
Of course I can move it to another Capabilities structure, please just let me know :)

PR #113, opened 2020-03-25
Add buildTarget/jvmEnvironment endpoint

lukaszwawrzyk:
This PR extends JvmEnvironmentItem with jdk (home uri and version string).
Also adds another endpoint jvmEnvironment (in addition to jvmTestEnvironment as generally this might be different.

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/JvmExtension.xtend
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/JvmExtension.xtend
@@ -6,6 +6,16 @@ import org.eclipse.lsp4j.jsonrpc.validation.NonNull
 import org.eclipse.lsp4j.generator.JsonRpcData
 
 @JsonRpcData
+class JvmBuildTarget {
+  String javaHome
+  String javaVersion
+  new(String javaHome, String javaVersion) {
+    this.javaHome = javaHome
+    this.javaVersion = javaVersion
+  }
+}
+
+@JsonRpcData
 class JvmTestEnvironmentParams {
   @NonNull List<BuildTargetIdentifier> targets
   new(@NonNull List<BuildTargetIdentifier> targets) {
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaExtension.xtend
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaExtension.xtend
@@ -12,13 +12,15 @@ class ScalaBuildTarget {
   @NonNull String scalaBinaryVersion
   @NonNull ScalaPlatform platform
   @NonNull List<String> jars
+  @NonNull JvmBuildTarget jvmBuildTarget
   new(@NonNull String scalaOrganization, @NonNull String scalaVersion, @NonNull String scalaBinaryVersion,
-      @NonNull ScalaPlatform platform, @NonNull List<String> jars) {
+      @NonNull ScalaPlatform platform, @NonNull List<String> jars, @NonNull JvmBuildTarget jvmBuildTarget) {
     this.scalaOrganization = scalaOrganization
     this.scalaVersion = scalaVersion
     this.scalaBinaryVersion = scalaBinaryVersion
     this.platform = platform
     this.jars = jars
+    this.jvmBuildTarget = jvmBuildTarget
   }
 }
 
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildTargetDataKind.java
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildTargetDataKind.java
@@ -1,6 +1,7 @@
 package ch.epfl.scala.bsp4j;
 
 public class BuildTargetDataKind {
+    public static final String JVM = "jvm";
     public static final String SCALA = "scala";
     public static final String SBT = "sbt";
 }
--- a/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/JvmBuildTarget.java
+++ b/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/JvmBuildTarget.java
@@ -1,0 +1,75 @@
+package ch.epfl.scala.bsp4j;
+
+import org.eclipse.xtext.xbase.lib.Pure;
+import org.eclipse.xtext.xbase.lib.util.ToStringBuilder;
+
+@SuppressWarnings("all")
+public class JvmBuildTarget {
+  private String javaHome;
+  
+  private String javaVersion;
+  
+  public JvmBuildTarget(final String javaHome, final String javaVersion) {
+    this.javaHome = javaHome;
+    this.javaVersion = javaVersion;
+  }
+  
+  @Pure
+  public String getJavaHome() {
+    return this.javaHome;
+  }
+  
+  public void setJavaHome(final String javaHome) {
+    this.javaHome = javaHome;
+  }
+  
+  @Pure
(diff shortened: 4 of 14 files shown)


bsp4j/src/main/java/ch/epfl/scala/bsp4j/JvmExtension.xtend
```diff
@@ -30,6 +24,30 @@ class JvmEnvironmentItem {
...
+  new(@NonNull List<BuildTargetIdentifier> targets) {
+    this.targets = targets
+  }
+}
+
+@JsonRpcData
+class JvmEnvironmentResult {
+  @NonNull List<JvmEnvironmentItem> items
+  new(@NonNull List<JvmEnvironmentItem> items) {
+    this.items = items
```
olafurpg: I don't think the `jvmEnvironment` should include a classpath and I think it can be per-workaspace.

```scala
case class JvmEnvironmentParams()
case class JvmEnvironmentResult(javaHome: String, javaVersion: String)
```

What do you think @jastice ?

lukaszwawrzyk: That was my initial implementation, yet Justin mentioned that in general jdk can be different per target. We will be able to use JvmEnvironment for running apps like we use JvmTestEnvironment for running tests.

olafurpg: That's fair. I don't think the response should include the classpath however, this method should be cheap to respond to. I wonder if it's also worth adding the ability to say the JDK is the same for all targets 🤔

lukaszwawrzyk: Hmm, maybe we should put jdk and other runtime parameters in separate endpoints. Not sure how to name the new one. It could respond with say:
```
case class JdkResponse(projectJdk: Option[JdkItem], moduleJdks: List[JdkItem])
case class JdkItem(buildTargetId: Option[BuildTargetId], javaHome: String, javaVersion: String)
```

jastice: Having the classpath in there has the benefit that we effectively get Java support for non-Scala modules

What's the reason this method in particular should be cheap? I'm not against splitting it to separate endpoints, just wondering in what cases it would be beneficial.

tpasternak:
Should we include 
`
  String javaHome
  String javaVersion
`
in `JvmTestEnvironmentResult` as well?

tpasternak:
Should we include 
`
  String javaHome
  String javaVersion
`
in `JvmTestEnvironmentResult` as well?

bsp4j/src/main/java/ch/epfl/scala/bsp4j/JvmExtension.xtend
olafurpg: I believe I'm misunderstanding the information needed in IntelliJ to configure the JDK. My understanding is that this issue is to avoid the manual prompt to configure the "Project JDK"

Screenshot 2020-03-19 at 17 41 07

Is the goal is to support non-Scala targets instead?

olafurpg:
I think it makes sense to include the Java Home/Version in `JvmTestEnvironmentResult`, but I'm not sure if that helps avoid the "Setup JDK" prompt after you finish importing a project

Screenshot 2020-03-19 at 17 41 07

bsp4j/src/main/java/ch/epfl/scala/bsp4j/JvmExtension.xtend
jastice: The prompt is due to a missing module JDK configuration and a bug in the code that generates the prompt. I'm fixing this without any changes to BSP.

Adding this change to BSP however allows us to configure the JDK explicitly from the server rather than best-guessing or manually configuring it on the client side, which can be necessary if there is a version incompatibility.

Supporting Java-only targets is an added benefit, but a rather important one.

olafurpg: >  I'm fixing this without any changes to BSP.

How does IntelliJ infer the default Java home? The Java version influences the presentation compiler to resolve symbols like `java.nio.file.Files`. We should make sure that IntelliJ doesn't for example expose JDK 11 symbols when the build is on JDK 8.

The `jvmTestEnvironment` endpoint is only called right before running tests so I'm not sure if it's a good candidate to configure the presentation compiler.

Would it make sense to include the Java home and java version in a `JavaBuildTarget` the same way we do for `ScalaBuildTarget`? https://build-server-protocol.github.io/docs/extensions/scala.html We can make `ScalaBuildTarget` extend `JavaBuildTarget`. That way we can add an explicit Java extension to the protocol.

jastice: By default I just take the most recent one available. That of course has the problem you mention: it might expose methods you don't want, or be downright incompatible, This endpoint addresses that case.

`JavaBuildTarget` would be an alternative. On the other hand, these parameters would also apply to other JVM languages. I don't have a strong opinion on this yet.

Calling it before tests doesn't mean we can't call it any other time we need it.

olafurpg: > Calling it before tests doesn't mean we can't call it any other time we need it.

How would that work when the source code doesn't compile successfully? `jvmTestEnvironment` only returns if the code compiles without errors.

olafurpg: The reason I think it should be cheap to query the java home/version is that 1) this version is cheap to produce and 2) we need this information for the presentation compiler before triggering "build project"

bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaExtension.xtend
```diff
@@ -12,13 +12,15 @@ class ScalaBuildTarget {
   @NonNull String scalaBinaryVersion
   @NonNull ScalaPlatform platform
   @NonNull List<String> jars
+  @NonNull JvmBuildTarget jvmBuildTarget
```
olafurpg: I like this design over `ScalaBuildTarget extends JvmBuildTarget`. Composition is better than inheritance in this case 👍

olafurpg:
I think `JvmBuildTarget` is an OK name since it's consistent with `ScalaBuildTarget`. We could call it `JvmInfo` but then I would expect it to also include the classpath and system properties.

jvican:
My point is `ScalaBuildTarget` is an actual build target because one can run build tasks on it (compile, run, etc). It doesn't look like this is the case for `JvmBuildTarget`, it's just a name for a class encapsulating some data.

olafurpg:
> Any reason we aren't adding the classpath right now? Or should we do that in a separate JavaBuildTarget?

@jastice this information is included in the response of `workspace/buildTargets`, which can be responded without library dependency resolution or compilation. If we include the classpath then `workspace/buildTargets` will have a very different purpose.

olafurpg:
@jvican the `ScalaBuildTarget` class is not a "build target" by a similar definition since it doesn't include the target URI

```json
case class JvmBuildTarget(..)
case class ScalaBuildTarget(jvmBuildTarget: JvmBuildTarget, scalaVersion: ...)
case class BuildTarget(id: Uri, ..., data: Option[Json]) // data:
```

Do you propose a different name or should we merge this as is?

jvican:
You're right about that. Let's merge this as is!

PR #115, opened 2020-04-01
JvmEnvironmentItem's classpath as URI-formatted String

tpasternak:


tpasternak:
Btw I noticed the `SbtBuildTarget`'s `classpath` field is not documented in docs/extensions/bsp.md. I can put some info there, but I'm not sure what it's doing, it's neither implemented in Bloop nor used in IntelliJ

https://github.com/build-server-protocol/build-server-protocol/blob/master/bsp4j/src/main/java/ch/epfl/scala/bsp4j/SbtExtension.xtend#L11

jastice:
> Btw I noticed the `SbtBuildTarget`'s `classpath` field is not documented in docs/extensions/bsp.md. I can put some info there, but I'm not sure what it's doing, it's neither implemented in Bloop nor used in IntelliJ

Good point, I think it can be removed?

PR #116, opened 2020-04-03
Add new `buildTarget/jvmRunEnvironment` endpoint

tpasternak:
Mirror endpoint to jvmTestEnvironment - used to execute main class
from an external tool in the same way as BSP server is doing it

Follow up of https://github.com/build-server-protocol/build-server-protocol/pull/107#discussion_r375346983


tpasternak:
I created this according to discussions on my previous PR (`jvmTestEnvironment`), but now I'm not sure whether these two should be kept separately. Will there be any difference between these two in bloop?

tpasternak:
@jvican @jastice do you think we can merge this?

PR #118, opened 2020-05-04
Change ScalaMainClass to carry environment variables

alekseiAlefirov:
The idea is having an ability for a user to provide environmet variables for starting a debug session.

Bloop Issue here: https://github.com/scalacenter/bloop/issues/1264

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaExtension.xtend
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaExtension.xtend
@@ -118,11 +118,10 @@ class ScalaMainClass {
   @NonNull @SerializedName("class") String className
   @NonNull List<String> arguments
   @NonNull List<String> jvmOptions
-  @NonNull List<String> environmentVariables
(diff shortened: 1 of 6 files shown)


tgodzik:
@jvican any chance we could merge this? It's just needed to be able to send environment variables from the user.

bsp4j/src/main/java/ch/epfl/scala/bsp4j/ScalaExtension.xtend
```diff
@@ -118,9 +118,11 @@ class ScalaMainClass {
   @NonNull @SerializedName("class") String className
   @NonNull List<String> arguments
   @NonNull List<String> jvmOptions
-  new(@NonNull String className, @NonNull List<String> arguments, @NonNull List<String> jvmOptions) {
+  @NonNull List<String> environmentVariables
```
jastice, line 121: I don't think we can allow non-null here unless we want to break compatibility. It doesn't seem necessary to break at this point.

alekseiAlefirov:
Thanks for the review @jvican , @jastice ! I've added changes, making the new field nullable and fixing the scala doc.

PR #130, opened 2020-09-03
Change JVM extension to require the JDK to be provided instead of the JRE

andrefmrocha:
It has been discovered that IntelliJ as a client requires the JDK to be provided instead of the JRE. It was also found that servers such as Bloop already provide this information. Since the JRE is insufficient to build the environment, me and @jastice are now proposing a change to the `JVMBuildTarget`, in order for it to return the JDK instead of the JRE


olafurpg (approved):
LGTM 👍

PR #133, opened 2020-09-08
Add `manual` build target tag

jastice:
Add `manual` build target tag. This is based on the same manual tag in Bazel.

It is used when targets are not supposed to be considered by default when triggering a full workspace build. Instead a build/test/etc action on these targets should be triggered manually by the user.

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -184,6 +184,11 @@ export namespace BuildTargetTag {
 
   /** Target should be ignored by IDEs. */
   export const NoIDE = "no-ide";
+
+  /** Actions on the target such as build and test should only be invoked manually
+   * and explicitly. For example, triggering a build on all targets in the workspace
+   * should by default not include this target. */
+  export const Manual = "manual";
 }
 
 export interface BuildTargetCapabilities {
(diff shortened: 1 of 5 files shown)


docs/specification.md
```diff
@@ -184,6 +184,11 @@ export namespace BuildTargetTag {
 
   /** Target should be ignored by IDEs. */
   export const NoIDE = "no-ide";
+
+  /** Actions on the target such as build and test should only be invoked manually
+   * and explicitly. For example, triggering a build on all targets in the workspace
+   * should by default not include this target. */
```
olafurpg, line 190: What do you think about adding a mention of this feature in Bazel to further explain the motivation for this feature?
```suggestion
   * should by default not include this target.
   *
   * The original motivation to add the "manual" tag comes from a similar functionality that exists in Bazel:
   *
   * > Targets with tags = ["manual"] will not be included in wildcard target patterns (..., :*, :all, etc.). You should specify such test targets with explicit target patterns on the command line if you want Bazel to build/test them.
   */
```

olafurpg (approved):
This is a good idea! LGTM 👍 Feel free to ignore the suggestion

jvican (approved):
LGTM 👍

docs/specification.md
jastice: Good idea though I'd leave out the Bazel specifics. I'll reword it a bit.

Issue #134, opened 2020-09-10
Add build/reload request


adpi2:
The ability to trigger a build server refresh is currently missing from the protocol.

This issue is a follow up of https://github.com/sbt/sbt/issues/5783. Adding a `build/reload request ` will ease the integration between IntelliJ and sbt, as well as IntelliJ and Bloop.

Basically,
- `workspace/buildTargets` is safe and returns the latest known state of the build
- `build/reload` asks the server to refresh the state of the build but it can fail, in which case the server might stop responding

A typical scenario would be:
- the user changes the build configuration file then clicks the `reload` button
- a `build/reload` request is sent by the client to the server
- the server reload the configuration
- after reloading successfully it replies with a success response
- the client sends a `workspace/buildTargets` request to ask the new list of build targets

jastice:
I am writing up the spec and protocol implementation right now!

Relevant spec parts:

```ts
export interface BuildServerCapabilities {
[...]
      /** Reloading the build state through workspace/reload is supported */
  canReload?: Boolean
```

```
### Reload request

The `reload` request instructs the build server to reload the build configuration.
This request should be supported by build tools that keep their state in memory.
If the `reload` request returns with an error, it's expected that other requests 
respond with the previously known "good" state.

Request:

- method: `workspace/reload`
- params: `null`

Response:

- result: `null`
- error: code and message in case an error happens during reload. For example, 
when the build configuration is invalid
```

Subject to change with reviews of course.

jastice:
to discuss: shoudl it be `workspace/reload` or `build/reload`

adpi2:
Ok great, this is good!

I like `build/reload` better than `workspace/reload` because it is part of the server lifetime category with:
- `build/initialize`
- `build/shutdown`
- `builld/exit`

adpi2:
I let you finish the implementation :)

PR #135, opened 2020-09-10
Introduce build/reload request

jastice:
The `reload` request instructs the build server to reload the build configuration.
This request should be supported by build tools that keep their state in memory.
If the `reload` request returns with an error, it's expected that other requests
respond with the previously known "good" state.

This request is motivated by tools such as sbt, which maintain an in-memory state of the build, where reloading the build on-demand is not considered practical.

Prompted by discussion in https://github.com/sbt/sbt/issues/5783
Solves #134 

cc @adpi2

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -372,7 +372,7 @@ export interface BuildServerCapabilities {
    * via method buildTarget/resources */
   resourcesProvider?: Boolean;
 
-  /** Reloading the build state through build/reload is supported */
+  /** Reloading the build state through workspace/reload is supported */
   canReload?: Boolean
 
   /** The server sends notifications to the client on build
@@ -413,24 +413,6 @@ Notification:
 ```ts
 export interface InitializedBuildParams {}
 ```
-
-### Reload request
-
-The `reload` request instructs the build server to reload the build configuration.
-This request should be supported by build tools that keep their state in memory.
-If the `reload` request returns with an error, it's expected that other requests 
-respond with the previously known "good" state.
-
-Request:
-
-- method: `build/reload`
-- params: `null`
-
-Response:
-
-- result: `null`
-- error: code and message in case an error happens during reload. For example, 
-when the build configuration is invalid
 
 #### Shutdown Build Request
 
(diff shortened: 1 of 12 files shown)


jastice:
Discussion point: should this be `build/reload` or `workspace/reload`?

`build/reload` may be better than `workspace/reload` because it is part of the server lifetime category with:
- `build/initialize`
- `build/shutdown`
- `builld/exit`

adpi2:
You are right @jvican. It makes sense since we have the `uri` field in `workspace/buildTargets`.

So I guess we should rename it `workspace/reload` and have an `uri` field in the params.

Sorry @jastice for misleading you on this.

jastice:
> So I guess we should rename it `workspace/reload` and have an `uri` field in the params.

True, though the workspace is implicitly provided via the working directory of the build server, so we probably don't need a `uri` field

adpi2:
I think we do because some build server does not "live" in the same "working directory". Bloop, for instance, can manage many workspaces and he wants to know which one to reload.

jastice:
> I think we do because some build server does not "live" in the same working directory. Bloop, for instance, can manage many workspaces and he wants to know which one to reload.

Yes, but the server that the clients speaks to is always started in the workspace root directory. Note we don't have any root uri parameter for `workspace/buildTargets` either.

adpi2:
Oh yes you are right. Thank you for the clarification.

adpi2:
I am ready to submit the implementation of `workspace/reload` in sbt as soon as this one is merged!

See https://github.com/sbt/sbt/pull/5838

docs/specification.md
```diff
@@ -411,6 +414,24 @@ Notification:
 export interface InitializedBuildParams {}
 ```
 
+### Reload request
+
+The `reload` request instructs the build server to reload the build configuration.
```
olafurpg, line 419: Normally, the request/notification descriptions explicitly say whether its sent from the client or server.

```suggestion
The `reload` request is sent from the client to instruct the build server to reload the build configuration.
```

olafurpg (approved):
LGTM 👍 I'm open for both `build/reload` and `workspace/reload`, the comment is just a nit, feel free to ignore

docs/specification.md
jastice: added it!

jastice:
updated to `workspace/reload`

jvican (approved):
LGTM, thanks! 👍

PR #139, opened 2020-09-28
Add support for remote debugging

pvid:
Extend `DebugSessionParamsDataKind` to support remote debugging.

Prerequisite of supporting remote debugging in bloop (and metals), see https://github.com/scalacenter/bloop/pull/1378


tgodzik:
Do we need something like:
```scala
@JsonCodec final case class ScalaAttach(
    port: Long,
   hostname: String
)
```

or is it all solved internally in DAP and the microsoft java server?

tgodzik:
> Do we need something like:
> 
> ```scala
> @JsonCodec final case class ScalaAttach(
>     port: Long,
>    hostname: String
> )
> ```
> 
> or is it all solved internally in DAP and the microsoft java server?

Ach nevermind, it seems to be handled by DAP itself.

tgodzik:
@jastice @olafurpg  what do you think about this change? It will enable to attach to remote also in VS Code.

jvican (approved):
Thanks for the contribution @pvid, LGTM 👍

lukaszwawrzyk (approved):

PR #140, opened 2020-10-06
sbt 1.4.0 has built-in support

bjaglin:

PR #141, opened 2020-10-07
update existing implementations in intro

bjaglin:
I missed that in https://github.com/build-server-protocol/build-server-protocol/pull/140!


jastice (approved):
Thank you! I updated capitalization of names.

PR #142, opened 2020-10-15
Update sbt-mdoc, docusaurus, and actions.

ckipp01:
I was looking for something on the website today that I knew existed and couldn't find it. It looks like the release CI has been failing for the last month, which explain why I couldn't find the `workspace/reload` stuff I was looking for. I'm not 100% sure why it's failing as the error isn't extremely descriptive. There does seem to be a json parsing issue, but the file it's reporting looks fine to me (sidebars.json). I do see some warnings about outdated stuff, so this is more of a stab at just updating the site generation stuff in hopes that it fixes the issue. If not, then at least this stuff is updated 🤷.

The first failure looks to be this one when @olafurpg moved the site release to GH actions, and the site release has failed ever since. 

https://github.com/build-server-protocol/build-server-protocol/runs/1103916239?check_suite_focus=true

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,9 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: olafurpg/setup-scala@v7
+        with:
+          fetch-depth: 0
+      - uses: olafurpg/setup-scala@v10
       - uses: olafurpg/setup-gpg@v2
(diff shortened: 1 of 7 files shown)


jastice (approved):
Thanks, LGTM!

olafurpg (approved):
LGTM 👍 Thank you!

olafurpg:
The website release worked \o/ New docs are live https://build-server-protocol.github.io/docs/specification.html#reload-request

Issue #143, opened 2020-10-19
jvmOptions support for test run


mandeburka:
In my day-to-day work I'm using separate local-only cofig file for running tests. It's really annoying I can't use it with vscode/metals setup. It appears that corresponding functionality is not implemented in BSP in the firs place. Any plans or issues with implementing such functionality?

PR #144, opened 2020-10-22
add jvm options to ScalaTestParams

mandeburka:
Reasoning behind this is described in https://github.com/build-server-protocol/build-server-protocol/issues/143
I've taken changes from https://github.com/build-server-protocol/build-server-protocol/pull/96 and made them on latest master + made jvmOptions nullable.

### DIFF
--- a/docs/extensions/scala.md
+++ b/docs/extensions/scala.md
@@ -55,10 +55,10 @@ export interface ScalaTestParams {
    * It is the result of `buildTarget/scalaTestClasses`. */
   testClasses?: ScalaTestClassesItem[];
 
-  /** The jvm options to run tests with. In case of 
-   * conflict (some jvm options from a higher level) this (a) and the other (b) options should be 
-   * concatenated. In case the same jvm option appears 
-   * in both (a) an (b) than the one from (a) takes precedence.
+  /** The JVM options to run tests with. In case JVM options 
+   * are also defined on the build server, they should be 
+   * concatenated with these. In case they conflict, these 
+   * should take precedence.
    */
   jvmOptions?: String[];
 }
--- a/docs/extensions/scala.md
+++ b/docs/extensions/scala.md
@@ -55,7 +55,11 @@ export interface ScalaTestParams {
    * It is the result of `buildTarget/scalaTestClasses`. */
   testClasses?: ScalaTestClassesItem[];
 
-  /** The jvm options to run tests with. */
+  /** The jvm options to run tests with. In case of 
+   * conflict (some jvm options from a higher level) this (a) and the other (b) options should be 
+   * concatenated. In case the same jvm option appears 
+   * in both (a) an (b) than the one from (a) takes precedence.
+   */
   jvmOptions?: String[];
 }
 ```
--- a/docs/extensions/scala.md
+++ b/docs/extensions/scala.md
@@ -55,7 +55,7 @@ export interface ScalaTestParams {
    * It is the result of `buildTarget/scalaTestClasses`. */
   testClasses?: ScalaTestClassesItem[];
 
-  /** The JVM options to run tests with. Replace any options 
+  /** The JVM options to run tests with. They replace any options 
    * that are defined by the build server if defined.
    */
   jvmOptions?: String[];
--- a/docs/extensions/scala.md
+++ b/docs/extensions/scala.md
@@ -54,6 +54,9 @@ export interface ScalaTestParams {
   /** The test classes to be run in this test execution.
    * It is the result of `buildTarget/scalaTestClasses`. */
   testClasses?: ScalaTestClassesItem[];
+
+  /** The jvm options to run tests with. */
+  jvmOptions?: String[];
 }
 ```
 
--- a/docs/extensions/scala.md
+++ b/docs/extensions/scala.md
@@ -55,10 +55,8 @@ export interface ScalaTestParams {
    * It is the result of `buildTarget/scalaTestClasses`. */
   testClasses?: ScalaTestClassesItem[];
 
-  /** The JVM options to run tests with. In case JVM options 
-   * are also defined on the build server, they should be 
-   * concatenated with these. In case they conflict, these 
-   * should take precedence.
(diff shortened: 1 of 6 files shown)


jastice:
Thanks! I'm not opposed to this change, but there's some question marks. Usually the jvmOptions for running tests are defined on the build tool side. Or it may run a test framework in-process. What are the intended semantics when passing jvmOptions from the client, when the server already has a set of options defined?

mandeburka:
In tests I need to define which typesafe config file to load. The only way to do it is via JVM properties. Bloop, for example, already accepts jvm options to run tests with internally. My ultimate goal to be able to pass jvm options from vs code -> scala metals -> bloop

mandeburka:
I would like to clarify another thing as well. DebugSessionParams can have "scala-test-suites". What is contents of the data field in that case? I initially thought it should be ScalaTestParams but metals and bloop encodes/decodes it at List[String].

docs/extensions/scala.md
```diff
@@ -54,6 +54,9 @@ export interface ScalaTestParams {
   /** The test classes to be run in this test execution.
    * It is the result of `buildTarget/scalaTestClasses`. */
   testClasses?: ScalaTestClassesItem[];
+
+  /** The jvm options to run tests with. */
```
olafurpg, line 58: Can you elaborate on what happens when this option conflicts with existing JVM options that are configured in the build? Should the two be concatenated or should this override?

olafurpg (changes requested):
Thank you for this contribution! The code changes look good, just a comment on the spec. I think it makes sense to support the use-case to configure JVM options from the BSP client.

docs/extensions/scala.md
mandeburka: Done

docs/extensions/scala.md
```diff
@@ -55,7 +55,11 @@ export interface ScalaTestParams {
    * It is the result of `buildTarget/scalaTestClasses`. */
   testClasses?: ScalaTestClassesItem[];
 
-  /** The jvm options to run tests with. */
+  /** The jvm options to run tests with. In case of 
+   * conflict (some jvm options from a higher level) this (a) and the other (b) options should be 
+   * concatenated. In case the same jvm option appears 
+   * in both (a) an (b) than the one from (a) takes precedence.
```
jastice, line 61: I find the description to be a little confusing, as to what "higher level" is here. Does this capture your meaning?

> The JVM options to run tests with. In case JVM options are also defined on the build server, they should be concatenated with these. In case they conflict, these should take precedence.

mandeburka: Yes, well written. Thanks!
Will update the comment.

jastice (approved):

docs/extensions/scala.md
```diff
@@ -54,6 +54,13 @@ export interface ScalaTestParams {
   /** The test classes to be run in this test execution.
    * It is the result of `buildTarget/scalaTestClasses`. */
   testClasses?: ScalaTestClassesItem[];
+
+  /** The JVM options to run tests with. In case JVM options
```
jvican, line 58: I don't think this is what we want because servers cannot find conflicts easily between jvm options. That operation requires parsing and sometimes even a semantic understanding of the JVM options, as some of them use keywords such as "Enable" or "Disable" which cancel each other.
  
I think the right semantics for this new addition should be that they override the JVM test options specified in the build.

jvican (changes requested):
Hey :wave: Overall looks good to me but I have an important remark about semantics.

docs/extensions/scala.md
jastice: isn't the semantics of concatenation already that options later in the list override earlier ones?

jvican: Concatenation doesn't solve the problem because there are JVM options that can't be enabled when other JVM options are and viceversa, so when the JVM finds that it will fail early and exit.

That's why only the JVM options passed by the user have to be used uniquely by the build server, there's no sensible merge strategy, that needs to be done by the user changing the JVM options.

jastice: Yup, you're right. Good catch.

mandeburka: OK, correct me if I'm wrong. The decision here is to state that jvm options defined by user will be overwritten by jvm options defined by build server is they collide, right?

jvican: No, the solution is that the semantics of this field needs to change completely. All and only the JVM options provided here will be used by the build server to test the application, there's no merge strategy with the ones the build server has.

mandeburka: OK, thanks for clarification.

docs/extensions/scala.md
```diff
@@ -55,10 +55,8 @@ export interface ScalaTestParams {
    * It is the result of `buildTarget/scalaTestClasses`. */
   testClasses?: ScalaTestClassesItem[];
 
-  /** The JVM options to run tests with. In case JVM options 
-   * are also defined on the build server, they should be 
-   * concatenated with these. In case they conflict, these 
-   * should take precedence.
+  /** The JVM options to run tests with. Shouldn't be 
+   * extended or overwritten by the build server.
```
jastice, line 59: phrasing suggestion:

> If the request specifies jvmOptions, they replace any options that are defined by the build server.

mandeburka: Improved

jastice (approved):

docs/extensions/scala.md
```diff
@@ -54,6 +54,11 @@ export interface ScalaTestParams {
   /** The test classes to be run in this test execution.
    * It is the result of `buildTarget/scalaTestClasses`. */
   testClasses?: ScalaTestClassesItem[];
+
+  /** The JVM options to run tests with. Replace any options
```
jvican, line 58:
```suggestion
  /** The JVM options to run tests with. They replace any options 
```

jvican (approved):
LGTM :+1:

olafurpg (approved):
LGTM 👍 The new semantics make things clearer.

PR #155, opened 2021-01-01
Improve testkit error readability

andrefmrocha:
Removed logging of cancellation expections coming cancelling the server after tests were finished

Improved readability of testkit assertion related errors, by removing the `ExecutionException` layer and only throwing the assertion error, which improves readability of errors from
```
java.util.concurrent.ExecutionException: Boxed Error
ch.epfl.scala.bsp.testkit.client.TestFailedException: java.util.concurrent.ExecutionException: Boxed Error
(...)
```

to 

```
java.lang.AssertionError: assertion failed: Cleaned cache successfully, when it should have failed
ch.epfl.scala.bsp.testkit.client.TestFailedException: java.lang.AssertionError: assertion failed: Cleaned cache successfully, when it should have failed
(...)
```

### DIFF
--- a/testkit/src/main/scala/ch/epfl/scala/bsp/testkit/client/mock/MockSession.scala
+++ b/testkit/src/main/scala/ch/epfl/scala/bsp/testkit/client/mock/MockSession.scala
@@ -2,9 +2,10 @@ package ch.epfl.scala.bsp.testkit.client.mock
 
 import java.io.{InputStream, OutputStream}
 import java.util.concurrent.{ExecutorService, Executors}
-
 import ch.epfl.scala.bsp4j.{BuildServer, InitializeBuildParams}
 import org.eclipse.lsp4j.jsonrpc.Launcher
+
+import scala.util.Try
 
 case class MockSession(
     in: java.io.InputStream,
@@ -28,7 +29,7 @@ case class MockSession(
       .create()
 
     val listening = launcher.startListening
-    new Thread (() => listening.get()).start()
+    new Thread (() => Try(listening.get())).start()
 
     val server: BuildServer = launcher.getRemoteProxy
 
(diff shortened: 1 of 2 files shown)


jastice (approved):

testkit/src/main/scala/ch/epfl/scala/bsp/testkit/client/mock/MockSession.scala
```diff
@@ -28,7 +29,7 @@ case class MockSession(
       .create()
 
     val listening = launcher.startListening
-    new Thread (() => listening.get()).start()
+    new Thread (() => Try(listening.get())).start()
```
jvican, line 32: Aren't we ignoring any exception thrown here? Why is this necessary?

jvican (approved):
Thanks for the contribution! LGTM overall 👍 I leave a question though

testkit/src/main/scala/ch/epfl/scala/bsp/testkit/client/mock/MockSession.scala
andrefmrocha: Any exception thrown here will automatically impact the result of a test, therefore I don't think handling them is necessary. It is, however, wrapped in a `Try` to prevent the output of the `CancellationException` whenever the connection is shutdown, since this exception is irrevelant for the user

jvican: Then the ignoring should be explicit. Let’s replace it with a try catch expression, it’s clearer

andrefmrocha: Makes sense, changed!

jvican (approved):

PR #156, opened 2021-01-22
Fix generators

andrefmrocha:
Fixes problems existing in the shrinkers of URIs, that caused them to launch several exceptions when used.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -90,7 +90,7 @@ lazy val `bsp-testkit` = project
     bashScriptExtraDefines += """addJava "-Dscript.path=${app_home}/"""" + executableScriptName.value,
     batScriptExtraDefines += """call :add_java "-Dscript.path=%APP_HOME%\\"""" + executableScriptName.value + ".bat",
     libraryDependencies ++= List(
-      "org.scalacheck" %% "scalacheck" % "1.14.0",
+      "org.scalacheck" %% "scalacheck" % "1.15.2",
       "org.scala-lang.modules" %% "scala-java8-compat" % "0.9.1",
       "org.scala-lang.modules" %% "scala-collection-compat" % "2.1.6"
     ),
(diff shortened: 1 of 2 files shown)


testkit/src/main/scala/ch/epfl/scala/bsp/testkit/gen/UtilShrinkers.scala
```diff
@@ -22,18 +23,25 @@ trait UtilShrinkers {
...
-      scheme <- shrink(uri.getScheme)
-      host <- shrink(uri.getHost)
-      port <- shrink(uri.getPort)
+      scheme <- shrinkUriPartToLetters.shrink(uri.getScheme)
+      host <- shrinkUriPartToLetters.shrink(uri.getHost)
+      port <- shrinkPortInt.shrink(uri.getPort)
       path <- shrinkUriPath.shrink(uri.getPath)
     } yield {
       val a = new URI(scheme, null, host, port, path, null, null)
-      val b = new URI(scheme, null, null, port, path, null, null)
```
andrefmrocha, line 34: This use case with a null host always throws an exception on initialization

build.sbt
```diff
@@ -90,7 +90,7 @@ lazy val `bsp-testkit` = project
     bashScriptExtraDefines += """addJava "-Dscript.path=${app_home}/"""" + executableScriptName.value,
     batScriptExtraDefines += """call :add_java "-Dscript.path=%APP_HOME%\\"""" + executableScriptName.value + ".bat",
     libraryDependencies ++= List(
-      "org.scalacheck" %% "scalacheck" % "1.14.0",
+      "org.scalacheck" %% "scalacheck" % "1.15.2",
```
andrefmrocha, line 93: The new version of scalacheck allows us to use the method `suchThat` that allows allows for a better control how the shrinking actually happens

andrefmrocha:
I left some comments in order to better explain the changes upfront! :smile:

jastice (approved):

PR #160, opened 2021-02-03
Add definition of a C++ extension

andrefmrocha:
In this PR I'm looking into defining the behaviour of C++ for BSP!


bsp4j/src/main/java/ch/epfl/scala/bsp4j/CppBuildServer.java
```diff
@@ -0,0 +1,11 @@
+package ch.epfl.scala.bsp4j;
+
+import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
+
+import java.util.concurrent.CompletableFuture;
+
+public interface CppBuildServer {
+
+    @JsonRequest("buildTarget/cppDependenciesSources")
+    CompletableFuture<CppDependenciesSourcesResult> cppDependenciesSources(CppDependenciesSourcesParams params);
```
andrefmrocha, line 10: I think there is probably a better name for this but I can't find them.
The goal of this is to provide a means of setting the initial path an include should have. In c++ build systems, such as bazel it is to common for the import path to depend on the name of the path since c++ relies on paths for imports. An import in bazel, for instance, requires the following:
```c++
#include "<package_name>/path_to_header"
```

bsp4j/src/main/java/ch/epfl/scala/bsp4j/CppExtension.xtend
```diff
@@ -0,0 +1,43 @@
...
+  @NonNull List<CppDependenciesSourcesItem> items
+  new(@NonNull List<CppDependenciesSourcesItem> items) {
+    this.items = items
+  }
+}
+
+@JsonRpcData
+class CppDependenciesSourcesItem {
+  @NonNull BuildTargetIdentifier target
+  String packageNamePrefix
```
andrefmrocha, line 39: Given that this request for now exists just for getting this prefix, should it be in the constructor? I'm  not very keen on doing that if there is ever a need for more data on it.

bsp4j/src/main/java/ch/epfl/scala/bsp4j/CppExtension.xtend
```diff
@@ -0,0 +1,43 @@
+package ch.epfl.scala.bsp4j
+
+import java.util.List
+import org.eclipse.lsp4j.jsonrpc.validation.NonNull
+import org.eclipse.lsp4j.generator.JsonRpcData
+
+@JsonRpcData
+class CppBuildTarget {
+  @NonNull CppPlatform platform
```
andrefmrocha, line 9: Should this be an enum or a string? A string is what is used for the java version, but that might be because that usually uses semantic versioning, whilst in c++ we have to handle cases such as `C++11`

jastice: Probably just a String, so that implementor-side extensions are easier and don't later conflict with the mainline protocol. We can have a list of predefined constants with agreed on meaning in the protocol.

andrefmrocha:
We've asked someone from the Clion to help us understand if this is a good specification, and they did not have anything to point out, so I think this is ready for review!

adpi2 (approved):
Thanks for this contribution. It is amazing to have C/C++ support in BSP :)

Sorry for taking so long to review.

Issue #145, opened 2020-10-26
[Proposal] - Can we add a `DebugAdaptorProvider` (or something similiar) into the BuildServerCapabilities?


ckipp01:
While working on sbt BSP integration in Metals, I came a across a use case that having something like `DebugAdaptorProvider` or something similar in `BuildServerCapabilities` would have been quite useful. I noticed that in the https://github.com/build-server-protocol/build-server-protocol/pull/135 a `canReload` was added which is useful for the client to check to see if the server implements it before requesting it. In Metals we generate code lenses to trigger debug runs, and after seeing the reload pr my instinct was to check to see if the build server was a DAP provider via checking the `BuildServerCapabilities` before generating them. However, looking at https://github.com/build-server-protocol/build-server-protocol/pull/99 it seems that wasn't part of the pr.

Would it be possible to adapt the spec a bit to include something like this for clients to ensure a build server implements it before sending in a `debugSession/start`?

tgodzik:
I agree this would be very useful, especially with more BSP servers being created.

jastice:
Yup, this shouldn't be a big problem to add, since there's only one implementation I think?

er1c:
I noticed that `debugSession/start`, while the messages are defined in code: https://github.com/build-server-protocol/build-server-protocol/blob/master/bsp4s/src/main/scala/ch/epfl/scala/bsp/endpoints/Endpoints.scala#L72  There is no mention of it in the specification.  Is this related, or a separate issue than this ticket?

er1c:
I added a PR for this, please review!

PR #161, opened 2021-02-15
Add debug provider to build server capabilities

er1c:
This attempts to address https://github.com/build-server-protocol/build-server-protocol/issues/145

Specifically, looking to improve how metals can support debug support: https://github.com/adpi2/metals/pull/1/files#diff-2249f2a7f1d0baedcc556404e0eb12adcd38f5fdb4d0d640d662643bc3de5b1fR514

TODO:
- [x] Add an explicit json deserialization test for `BuildTargetCapabilities` to verify backward compatability

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
@@ -32,11 +32,13 @@ class BuildTargetCapabilities {
   @NonNull Boolean canCompile
   @NonNull Boolean canTest
   @NonNull Boolean canRun
+  @NonNull Boolean canDebug
 
-  new (Boolean canCompile, Boolean canTest, Boolean canRun) {
+  new (Boolean canCompile, Boolean canTest, Boolean canRun, Boolean canDebug) {
     this.canCompile = canCompile
     this.canTest = canTest
     this.canRun = canRun
+    this.canDebug = canDebug
   }
 }
 
@@ -81,7 +83,6 @@ class InitializeBuildParams {
   }
 }
 
-
 @JsonRpcData
 class BuildClientCapabilities {
   @NonNull List<String> languageIds
@@ -89,7 +90,6 @@ class BuildClientCapabilities {
     this.languageIds = languageIds
   }
 }
-
 
 @JsonRpcData
 class CompileProvider {
@@ -116,10 +116,19 @@ class RunProvider {
 }
 
 @JsonRpcData
+class DebugProvider {
+  @NonNull List<String> languageIds
+  new(@NonNull List<String> languageIds) {
+    this.languageIds = languageIds
+  }
+}
+
+@JsonRpcData
 class BuildServerCapabilities {
   CompileProvider compileProvider
   TestProvider testProvider
   RunProvider runProvider
+  DebugProvider debugProvider
   Boolean inverseSourcesProvider
   Boolean dependencySourcesProvider
   Boolean resourcesProvider
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
@@ -32,7 +32,7 @@ class BuildTargetCapabilities {
   @NonNull Boolean canCompile
   @NonNull Boolean canTest
   @NonNull Boolean canRun
-  @NonNull Boolean canDebug
+  Boolean canDebug = false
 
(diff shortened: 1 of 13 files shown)


olafurpg (changes requested):
Thank you for this contribution! This is a good improvement. The implementation looks good. Can you please update `specification.md` to match the implementation?

For better or worse, there's a lot of repetition in adding new BSP features 😅

er1c:
@olafurpg thanks, I've attempted to add the correct information to `specification.md`

olafurpg (approved):
LGTM 👍 thank you! cc/ @tgodzik does this look ok from Metals perspective?

docs/specification.md
```diff
@@ -1317,6 +1326,42 @@ response.
 The client will get a `originId` field in `RunResult` if the `originId` field in
 the `RunParams` is defined.
 
+### Debug Request
+
+The debug request is sent from the client to the server to debug build target(s). The
+server launches a [Microsoft DAP for Java](https://github.com/microsoft/vscode-java-debug) instance
```
adpi2, line 1332: We should keep this part language agnostic if possible. So I think it's better to link to the DAP specifiaction rather than to the Java implementation.

```suggestion
server launches a [Microsoft DAP](https://microsoft.github.io/debug-adapter-protocol/) server
```

tgodzik:
I think we should already be able to use canRun/canDebug - I didn't realise we had that. And since we implement both that and debug I don't imagine it would be super useful in case of Metals. I do believe it's a worthwhile addition and might be useful in the future.

er1c:
> I think we should already be able to use canRun/canDebug - I didn't realise we had that. And since we implement both that and debug I don't imagine it would be super useful in case of Metals. I do believe it's a worthwhile addition and might be useful in the future.

FYI: `canDebug` is only added/available as part of this PR, so I'm not sure I understand your comment about it not being useful to Metals.

adpi2 (approved):
LGTM

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
```diff
@@ -32,11 +32,13 @@ class BuildTargetCapabilities {
   @NonNull Boolean canCompile
   @NonNull Boolean canTest
   @NonNull Boolean canRun
+  @NonNull Boolean canDebug
```
jastice, line 35: making this `@NonNull` will make this change non-backward compatible

jastice (changes requested):

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
er1c: removed this and regenerated `xtend`

jastice: Thanks! I think in the Scala version it would need to be `Option[Boolean]` also, even though that is kind of ugly ...

er1c: I'm not super familiar with the syntax, but is there a way to make it a default value false?

jastice: you could do `canDebug: Boolean = false` but I'm not sure if/how that interacts with the serialization? I guess if that passes tests it's fine ;)

tests/src/test/scala/tests/SerializationSuite.scala
```diff
@@ -75,4 +75,26 @@ class SerializationSuite extends FunSuite {
...
   }
+
+  test("BuildTargetCapabilities - backward compatible canDebug") {
+    val legacyJson =
+      """
+        |{
+        |  "canCompile": true,
+        |  "canTest": true,
+        |  "canRun": true,
+        |  "canDebug": false
```
er1c, line 86: F**** me, no wonder it passed :)

tests/src/test/scala/tests/SerializationSuite.scala
```diff
@@ -11,8 +11,7 @@ import org.scalatest.FunSuite
 import scala.collection.JavaConverters._
 
 class SerializationSuite extends FunSuite {
-
-  val gson: Gson = new GsonBuilder().setPrettyPrinting().create()
+  val gson: Gson = new GsonBuilder().serializeNulls().setPrettyPrinting().create()
```
er1c, line 14: oops revert

jastice (approved):
Thanks for the compat fixes!

Issue #164, opened 2021-03-08
[Proposal] Add a new method for retrieving dependency modules


dos65:
### Motivation

Currently, BSP clients have a way to get dependencies and source artifacts: `ScalaTarget.classpath` / `DependencySourcesItem.sources`.
However, these fields have`Array[Uri/String]` type and there is no way to get a meta-information from them like `organization`, `artifact-name`, `version`.

For example, in the case of having this information, Metals might represent dependency sources in a more convenient way for users. Instead of extracting  them directly into `.readonly` it might add an additional root directory :`.readonly/$org:$artifact$version`. Also, this thing would help internally for indexing purposes.

### Proposal

Add an additional optional method to BSP - `buildTarget/dependencyModules` that returns a list of dependency modules with attached meta information. 

The sample structure might be the following one (I've copied it from bloop configuration format):
```ts
export interface DependencyModulesResult {
  items: DependencyModuleItem[];
}

export interface DependencyModuleItem {
  target: BuildTargetIdentifier;
  moduels: DependencyModule[];
}

export interface DependencyModule {
   organization: String;
   name: String;
   version: String;
   artifacts: Artifact[] ;
}

export interface Artifact {
   path: Uri;
   // `sources`/`docs` or empty
   classifier?: String;
}

```

jastice:
I think this would be a reasonable addition to the JVM extension

jvican:
Fixed in #165

PR #165, opened 2021-03-16
Add `buildTarget/dependencyModules` BSP method

dos65:
Based on #164 proposal. 

Adds the new BSP method - `buildTarget/dependencyModules`
It's a kind of extended version of  `buildTarget/sources` that has attached metadata about dependency libraries.

The only difference with the original proposal is that all JVM-specific data (org, classifiers) from DependencyModule is moved into the JVM extension.

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/MavenExtension.xtend
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/MavenExtension.xtend
@@ -1,0 +1,35 @@
+package ch.epfl.scala.bsp4j
+
+import java.util.List
+import org.eclipse.lsp4j.jsonrpc.validation.NonNull
+import org.eclipse.lsp4j.generator.JsonRpcData
+
+@JsonRpcData
+class MavenDependencyModule {
+  @NonNull String organization
+  @NonNull String name
+  @NonNull String version
+  @NonNull List<MavenDependencyModuleArtifact> artifacts
+  String scope
+  new(
+    @NonNull String organization,
+    @NonNull String name,
+    @NonNull String version,
+    @NonNull List<MavenDependencyModuleArtifact> artifacts
+  ) {
+    this.organization = organization
+    this.name = name
+    this.version = version
+    this.artifacts = artifacts
+  }
+}
+
+@JsonRpcData
+class MavenDependencyModuleArtifact {
+  @NonNull String uri
+  String classifier
+
+  new (@NonNull String uri) {
+    this.uri = uri
+  }
+}
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
@@ -145,6 +145,7 @@ class BuildServerCapabilities {
   DebugProvider debugProvider
   Boolean inverseSourcesProvider
   Boolean dependencySourcesProvider
+  Boolean dependencyModulesProvider
   Boolean resourcesProvider
   Boolean buildTargetChangedProvider
   Boolean jvmRunEnvironmentProvider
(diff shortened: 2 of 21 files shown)


docs/extensions/jvm.md
```diff
@@ -101,3 +101,31 @@ export interface JvmRunEnvironmentResult{
     environmentVariables: Map<String, String>;
 }
 ```
+
+### JVM Dependency Module
```
adpi2, line 105: I think this extension should be called `Maven Dependency Module` because:
- there are other kinds of dependencies with slightly different concepts (Ivy contains `extensions` but Maven does not)
- Maven Dependency Modules can also be used for Scala.js

docs/extensions/jvm.md
```diff
@@ -101,3 +101,31 @@ export interface JvmRunEnvironmentResult{
...
+
+`JvmDependencyModule` is a basic data structure that contains jvm-specific
+metadata. This metadata is embedded in the `data: Option[Json]` field of the `DependencyModule` definition, when the `dataKind` field contains "jvm".
+
+```ts
+export interface JvmDependencyModule {
+  organization: String;
+  name: String;
+  version: String;
+
```
adpi2, line 115: If it is a `Maven` dependency then it should also contain a optional `scope?: String`
```suggestion
scope?: String

```
Most of the time `scope` can be left empty, meaning it is a dependency on the `"compile"` scope.

adpi2 (commented):
Thank you for this contribution! It's a valuable addition to BSP.

My only concern is that `JVMDependencyModule` is not really specific to the JVM. I think it should be called `MavenDependencyModule`.

docs/extensions/jvm.md
dos65: I expected that there will be no difference between Maven/Ivy dependencies.
It's more like `ModuleID` + resolution data in sbt terms.

adpi2: sbt `ModuleId` looks simple but it is more complex than Maven or Ivy modules: https://github.com/sbt/librarymanagement/blob/23b7d923af2e93d5eebd67c93352758160462388/core/src/main/contraband/librarymanagement.json#L335-L353

adpi2: I want to call it `MavenModuleId` because it is the simplest one that is used in the JVM ecosystem. It does not imply that all dependencies are coming from Maven repositories.
Then if we want to add support for `Scala.js` we can reuse it.We can also add a `NpmModuleId` for native js libraries.

dos65: > It does not imply that all dependencies are coming from Maven repositories.

Oh ok, I got what you mean. Will rename it!

docs/extensions/jvm.md
dos65: Thx! Added

docs/extensions/jvm.md
dos65: Done

bsp4j/src/main/java/ch/epfl/scala/bsp4j/MavenExtension.xtend
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List
+import org.eclipse.lsp4j.jsonrpc.validation.NonNull
+import org.eclipse.lsp4j.generator.JsonRpcData
+
+@JsonRpcData
+class MavenDependencyModule {
+  @NonNull String organization
+  @NonNull String name
+  @NonNull String version
+  @NonNull List<MavenDependencyModuleArtifact> artifacts
```
adpi2, line 12: For completeness we should have `scope` here

bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
```diff
@@ -358,6 +359,42 @@ object SourceItemKind {
...
+  uri: String,
+  classifier: Option[String]
+)
+
+@JsonCodec final case class MavenDependencyModule(
+  organization: String,
+  name: String,
+  version: String,
+  artifacts: List[MavenDependencyModuleArtifact]
+)
```
adpi2, line 396: We should have `scope` here as well

bsp4j/src/main/java/ch/epfl/scala/bsp4j/MavenExtension.xtend
dos65: Ough! Thx - fixed!

bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
dos65: Added!

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
```diff
@@ -641,3 +642,45 @@ class DebugSessionAddress {
...
+class DependencyModule {
+  @NonNull String name
+  @NonNull String version
+  String dataKind
+  @JsonAdapter(JsonElementTypeAdapter.Factory) Object data
+
+  new (@NonNull String name, @NonNull String version) {
+    this.name = name
+    this.version = version
+    this.dataKind = dataKind
```
adpi2, line 684: This line is harmless but useless.

adpi2 (approved):
LGTM!

I am letting it open for some more days so that @jastice and @olafurpg have time to take a look at it if they want to.

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
dos65: removed

jastice (approved):
Thank you, looks good to me!

PR #167, opened 2021-03-29
Remove classpath from sbt build target

andrefmrocha:
The classpath had previously been removed from the `SbtBuildTarget` but it was still available through the constructor of the java version.


jastice (approved):

jastice:
Not sure if this affects library compatibility. Any idea @adpi2 ?

adpi2 (approved):
I think we can deal with the compatibility breakage:
- sbt does not yet support the `SbtExtension`
- Mill does not need it
- Metals does not consume the `classpath` in `SbtBuildTarget`
- Bloop is still using its own bsp library...

Issue #153, opened 2020-12-19
Cross Publish bsp4s


er1c:
I believe only 2.12 is published a 2.13 would be good.

ckipp01:
I actually just looked at this for something I was playing around with and thought I could quickly do this. However, there are a few snags here which I think may need to be figured out first.

1.  bs4s relies on lsp4s, which is archived in scalameta. From my understanding it was simply not worth maintaining another lsp library when lsp4j could just be used. So this isn't available for 2.13. This leaves us with a couple options
2a. We can use [jsonrpc4s](https://github.com/jvican/jsonrpc4s), which I see has 2.13 in the build file, and it's also included in https://github.com/build-server-protocol/build-server-protocol/pull/125/files
2b. Instead just use lsp4j instead. However, this probably changes quite a bit, and then you start to get into questions about what the main purpose of bsp4s is for that bsp4j can't provide. I don't know that answer to that.

I'd be interested in diving into this if one of the maintainers maybe has some insight into which directly they'd like to go with this.

jastice:
So far we've only maintained bsp4s because it's what bloop uses, I believe. lsp4s was too heavy on dependencies to be the primary bsp library even for Scala, which is why we also switched away from it in the IntelliJ implementation.

bsp4s offers a more Scala-like API than bsp4j with slightly better type representation of some concepts (field-specific types, Option instead of null, Scala Future instead of CompletableFuture) but otherwise functionally equivalent.

jvican:
Indeed, https://github.com/build-server-protocol/build-server-protocol/pull/125 brings support for my fork jsonrpc4s which is more lightweight on the dependency side and uses jsoniter-scala. I would like to keep supporting the Scala bsp4s side as it is. I can cross-compile jsonrpc4s to support 2.13 and Scala 3 as well. However, some help updating that PR would be welcome. @ckipp01 If you're up to it, I encourage you to rebase and push commits directly to the open PR to sync it with the latest master branch.

PR #174, opened 2021-05-26
Move to unified slash syntax for build.

ckipp01:
As of sbt 1.5.0 the old style syntax for sbt [has been
deprecated](https://eed3si9n.com/sbt-1.5.0). This fixes all the warning
when you start up sbt.

I also made a slight change to the `inferJavaHome()` method to ensure
that this compile as expected with both jdk 8 and 11.

I was looking at #153 and hit on these, so I figured I'd open a small separate pr for them.


adpi2 (approved):
Thanks

jastice (approved):

PR #125, opened 2020-07-27
Upgrade bsp4s to Monix 3.2.2 and use jsoniter-scala

jvican:
Removes dependency on circe, we now only depend on jsoniter-scala itself
for JSON purposes, which is a library with no third-party dependencies.

PR #175, opened 2021-05-27
Upgrade bsp4s to use jsonrpc4s and jsoniter-scala directly

ckipp01:
This is a continuation of the work that @jvican did in #125.

> Removes dependency on circe, we now only depend on jsoniter-scala itself
for JSON purposes, which is a library with no third-party dependencies.

I rebased on the current main branch and brought this up to date with the aim
to unlock #153.

~Note that that also includes the work in #174, so you may want to merge that in first.~

### DIFF
--- a/bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
+++ b/bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
@@ -2,11 +2,14 @@ package ch.epfl.scala.bsp
 
 import java.net.{URI, URISyntaxException}
 
-import io.circe.Decoder.Result
-import io.circe._
-import io.circe.derivation.JsonCodec
-import io.circe.generic.extras._
-import io.circe.generic.extras.semiauto.{deriveDecoder, deriveEncoder}
+import ch.epfl.scala.bsp.BuildTargetEventKind.{Changed, Created, Deleted}
+import com.github.plokhotnyuk.jsoniter_scala.core.JsonValueCodec
+import com.github.plokhotnyuk.jsoniter_scala.macros.JsonCodecMaker
+import com.github.plokhotnyuk.jsoniter_scala.core.JsonCodec
+import com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter
+import com.github.plokhotnyuk.jsoniter_scala.core.JsonReader
+import com.github.plokhotnyuk.jsoniter_scala.macros.CodecMakerConfig
+import jsonrpc4s.RawJson
 
 final case class Uri private[Uri] (val value: String) {
   def toPath: java.nio.file.Path =
@@ -17,36 +20,37 @@ object Uri {
   // This is the only valid way to create a URI
   def apply(u: URI): Uri = Uri(u.toString)
 
-  implicit val uriEncoder: RootEncoder[Uri] = new RootEncoder[Uri] {
-    override def apply(a: Uri): Json = Json.fromString(a.value)
-  }
-
-  implicit val uriDecoder: Decoder[Uri] = new Decoder[Uri] {
-    // Should we add validation here
-    override def apply(c: HCursor): Result[Uri] = {
-      c.as[String].flatMap { str =>
-        try Right(Uri(URI.create(str).toString))
-        catch {
-          case _: IllegalArgumentException | _: URISyntaxException =>
-            Left(DecodingFailure(s"String $str is not a valid URI.", c.history))
-        }
+  implicit val uriCodec: JsonValueCodec[Uri] = new JsonValueCodec[Uri] {
+    def nullValue: Uri = null
+    def encodeValue(id: Uri, out: JsonWriter): Unit = out.writeVal(id.value)
+    def decodeValue(in: JsonReader, default: Uri): Uri = {
+      val defaultStr = if (default == null) null else default.value
+      val str = in.readString(defaultStr)
+      try Uri(URI.create(str).toString)
+      catch {
+        case _: IllegalArgumentException | _: URISyntaxException =>
+          in.decodeError(s"String $str is not a valid URI!")
       }
     }
   }
 }
 
-@JsonCodec final case class TextDocumentIdentifier(
+final case class TextDocumentIdentifier(
     uri: Uri
 )
 
-@JsonCodec final case class BuildTargetIdentifier(
+object TextDocumentIdentifier {
+  implicit val codec: JsonValueCodec[TextDocumentIdentifier] =
+    JsonCodecMaker.make(CodecMakerConfig.withTransientEmpty(true))
+}
+
+final case class BuildTargetIdentifier(
     uri: Uri
 )
 
-object BuildTargetCapabilities {
-  implicit val config: Configuration = Configuration.default.withDefaults
-  implicit val jsonDecoder: Decoder[BuildTargetCapabilities] = deriveDecoder
-  implicit val jsonEncoder: Encoder[BuildTargetCapabilities] = deriveEncoder
+object BuildTargetIdentifier {
+  implicit val codec: JsonValueCodec[BuildTargetIdentifier] =
+    JsonCodecMaker.make(CodecMakerConfig.withTransientEmpty(true))
 }
 
 final case class BuildTargetCapabilities(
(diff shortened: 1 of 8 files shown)


bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
```diff
@@ -627,160 +932,300 @@ object ScalaPlatform {
...
     argv: List[String],
     version: String,
     bspVersion: String,
     languages: List[String]
 )
 
-@JsonCodec final case class DebugSessionParams(
+object BspConnectionDetails {
+  implicit val codec: JsonValueCodec[BspConnectionDetails] =
+    JsonCodecMaker.make(CodecMakerConfig.withTransientEmpty(false))
```
ckipp01, line 1133: So quick question on this @jvican. Originally you had `withTransientEmpty(true)` for all of these, but I want to verify that it was intentional? When testing, I found that with that, the behavior actually changes from how it was with circe. For example in the `SerializationPropertySuite` you can see that with the old tests and with this set to `false` you'll get something like this where the empty collections are encoded and still there as empty. 
```scala
SerializationPropertySuite.scala:30 bsp4sValueDecoded: BspConnectionDetails("", List(), "", "", List())
SerializationPropertySuite.scala:32 bsp4sJson: "{\"name\":\"\",\"argv\":[],\"version\":\"\",\"bspVersion\":\"\",\"languages\":[]}"
```

However, with `withTransientEmpty` set to `true`, the behavior changes where the empty collections are left out.
```scala
SerializationPropertySuite.scala:30 bsp4sValueDecoded: BspConnectionDetails("", List(), "", "", List())
SerializationPropertySuite.scala:32 bsp4sJson: "{\"name\":\"\",\"version\":\"\",\"bspVersion\":\"\"}"
```

Was this an intentional change that we want to change, or can I keep these as `false` to retain the behavior from before?

tests/src/test/scala/tests/TypoSuite.scala
```diff
@@ -194,21 +194,27 @@ class TypoSuite extends FunSuite {
...
   implicit class XtensionServices(services: Services) {
-    def forwardRequest[A, B](endpoint: Endpoint[A, B])(implicit client: JsonRpcClient): Services = {
-      services.requestAsync(endpoint)(a => endpoint.request(a))
+    def forwardRequest[A, B](endpoint: Endpoint[A, B])(implicit client: RpcClient): Services = {
+      //services.requestAsync(endpoint)(a => endpoint.request(a))
+
+      // FIXME  returns a Task[RpcReponse[B]] but needs a Task[B]
+      // Not fully sure how to fix this one
+
+      //services.requestAsync(endpoint)(a => endpoint.request(a))
```
ckipp01, line 203: Also this is the only thing I has left which is tripping me up. `endpoint.request()` here returns a `Task[RpcResponse[B]]` but a `Task[B]` is actually needed. I'm struggling with getting that to that point. Any pointers about what I'm missing here would be helpful.

jvican: Yeah, this is a bit unintuitive if you don't know about the underlying jsonrpc changes. I've added a code suggestion to handle this case, I'm not sure it works but it should.

tests/src/test/scala/tests/TypoSuite.scala
```diff
@@ -194,21 +194,27 @@ class TypoSuite extends FunSuite {
...
+      ???
```
jvican, line 204:
```suggestion
      services.requestAsync(endpoint) { a =>
        endpoint.request(a).flatMap {
          case RpcResponse.RpcSuccess(value, _) => Task.now(value)
          case err: RpcResponse.RpcFailure      => Task.raiseError(err)
        }
      }
```

jvican (approved):
Excellent :+1: Thanks for working on this. I've left a few comments, we should just do those minor changes and we should be good to go.

bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
ckipp01: Ahh ok cool, I've gone ahead and switched these all to use `makeWithRequiredCollectionFields` then 👍🏼

website/i18n/en.json
```diff
@@ -12,6 +12,10 @@
       "bsp": {
         "title": "bsp"
       },
+      "extensions/cpp": {
```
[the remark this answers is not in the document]
ckipp01: Ah, it wasn't intentional, but when I updated mdoc and tested it locally to ensure it worked, it must have created it. For now I've gone ahead and just removed it.

tests/src/test/scala/tests/TypoSuite.scala
ckipp01: 🤔 so I'm still struggling with this a bit. While this compiles find I can't figure out why this keeps timing out.

https://github.com/build-server-protocol/build-server-protocol/runs/2700373516?check_suite_focus=true#step:7:81

Any ideas?

jvican: I'm trying to look at this tomorrow, sorry for the delay @ckipp01

jvican: I'm so sorry Chris, as much as I try I can't get to take a look at this. Have you looked at the implementation of the tests and maybe even [jsonrpc4s](https://github.com/jvican/jsonrpc4s)? I don't think this should time out so maybe some exception is not being thrown properly or the tests wait on a promise that is never completed?

ckipp01: Hey, no worries at all. When I get a bit of free time I'll do a deeper dive into jsonrpc4s to see what's going on. From looking at the tests in here it wasn't really clear why this would happen, but I could totally be missing something that has changed between the lsp4s and jsonrpc4s implementation. I'll dig deeper.

build.sbt
```diff
@@ -32,12 +49,9 @@ lazy val bsp4s = project
   .settings(
     Test / publishArtifact := false,
     Compile / doc / sources := Nil,
-    addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full),
     libraryDependencies ++= List(
-      "io.circe" %% "circe-core" % "0.9.0",
-      "io.circe" %% "circe-derivation" % "0.9.0-M4",
-      "io.circe" %% "circe-generic-extras" % "0.9.0",
-      "org.scalameta" %% "lsp4s" % "0.2.0"
+      "me.vican.jorge" %% "jsonrpc4s" % "ae3705e0+20210611-1304",
```
ckipp01, line 53: - [ ] Update after jsonrpc4s is released.

This is a local version that I was working with. I made a pull request here: https://github.com/jvican/jsonrpc4s/pull/3 that can hopefully be merged and released. The current released version is missing some commits and differs a bit from what's on the master branch there atm. I couldn't get tests to pass without using those commits.

bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
```diff
@@ -75,54 +84,131 @@ object BuildTargetTag {
...
+        }
+
+      override def encodeValue(x: A, out: JsonWriter): Unit =
+        if (x == null) out.writeNull()
+        else codec.encodeValue(x, out)
+
+      override val nullValue: A = null.asInstanceOf[A]
+    }
+
+}
```
ckipp01, line 152: This is one of the other things I needed to add to make the requests work that allowed for `null`, for example `Reload`. This is also a bit more generic just to allow for it also be used in the case where you actually want to optionally allow for `null`. However in our case the main thing we are using it for is to just ensure capture the null value. Thanks to Andriy on Gitter for helping me figure this out.

ckipp01:
Alrighty, after updating to the newly released jsonrpc4s we are all 🟢.

jvican (commented):
We're golden 👌 This is good to go, we can merge after Justin's review.

jvican (approved):
We're golden 👌 This is good to go, we can merge after Justin's review.

jvican:
@jastice Can you take a look at this whenever you have some time?

jastice (approved):
LGTM! Thanks!

Issue #105, opened 2019-11-14
Update to the latest lsp4j dependency


olafurpg:

forty:
Hello :wave: Anything blocking for this? After going dependency fishing, I ended up here. Apparently lsp4j 0.8 causes an old version of Guava to be pulled, which breaks my completely unrelated mill build :)

ckipp01:
This is up to date now and can be closed.

PR #176, opened 2021-06-12
Update lsp4j to 0.9.0

forty:
I don't really know what I'm doing, but I need a more recent version of guava, and that seems to do the trick. `sbt test` succeeds. I did not update to a more recent version as starting from lsp4j 0.10.0 the test fails (and I have not idea how to fix them)


forty:
I actually found how to fix the tests. Once again, I have no idea what I'm doing, so please have this reviewed by some who do :)

This solves https://github.com/build-server-protocol/build-server-protocol/issues/105

jastice (approved):
Thanks for this!

PR #183, opened 2021-06-24
use Java 11 in github actions

jastice:
Fixes failing builds due to incompatible class files in xtend generation

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,9 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - uses: olafurpg/setup-scala@v10
+      - uses: olafurpg/setup-scala@v11
+        with:
+          java-version: adopt@1.11
       - uses: olafurpg/setup-gpg@v3
       - name: Publish ${{ github.ref }}
         run: sbt ci-release docs/docusaurusPublishGhpages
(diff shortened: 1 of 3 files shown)


adpi2:
If we use Java 11, will it be still possible to use bsp4 and bsp4j with Java 8?

tgodzik:
> If we use Java 11, will it still be possible to use bsp4 and bsp4j with Java 8?

I think we wouldn't be able to use Java 8, so I would hesitate to merge this :thinking:  We could instead add a separate verify job for JDK 11?

jastice:
I'm trying to figure out how these even broke. #176 would be a good candidate, but reverting to before that commit doesn't fix it. Not even to old 2020 commits. So what is actually going on here?

Meanwhile, we could use JDK11 only for the verify job but not release?

tgodzik:
> I'm trying to figure out how these even broke. #176 would be a good candidate, but reverting to before that commit doesn't fix it. Not even to old 2020 commits. So what is actually going on here?
> 
> Meanwhile, we could use JDK11 only for the verify job but not release?

Probably both jdk 11 and 8 are needed just in case, otherwise the release might fail even if check runs ok.

olafurpg (commented):
Can we add a build matrix for the verify job? The release job should continue to run on 8.

jastice:
obsoleted by #185

PR #180, opened 2021-06-17
Cross build all Scala modules for 2.12.x and 2.13.x

ckipp01:
**NOTE** ~this is built on top of the work in https://github.com/build-server-protocol/build-server-protocol/pull/175, so you'll want to merge that in first. The only real changes are in this commit: https://github.com/build-server-protocol/build-server-protocol/pull/180/commits/e20a92adbbc875ccdd4c9946eac4f7d3ce45d53a~

Closes #153

PR #185, opened 2021-06-27
Update xtend and add java matrix to build

ckipp01:
This is an attempt to fix the issues I'm seeing in #180. I also have no idea why this popped up since trying to do a git bisect on the changes seems to have everything failing and I can't find a good commit (like you mentioned in #183). After looking at what you have in #183 I decided to add in a matrix here for 8 and 11 to verify on and then changed the way we do `xtend` a bit. I modeled it after the way it's actually done in here: https://github.com/xuwei-k/sbtend/blob/master/src/main/scala/Plugin.scala. It seems to be working now for 8 and 11.

EDIT: Cool, all 🟢. @jastice this should take care of the issues we are seeing in #180 and #183.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -5,16 +5,28 @@ on:
   pull_request:
     branches: [master]
 jobs:
-  verify:
+  generate-and-test:
+    name: on ${{ matrix.java }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [adopt@1.8, adopt@1.11]
     steps:
       - uses: actions/checkout@v2
-      - uses: olafurpg/setup-scala@v10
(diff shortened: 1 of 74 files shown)


jastice (approved):
Thanks! Really wonder what happened with xtend there ...

.github/workflows/verify.yml
```diff
@@ -5,16 +5,28 @@ on:
...
+  generate-and-test:
+    name: on ${{ matrix.java }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [adopt@1.8, adopt@1.11]
     steps:
       - uses: actions/checkout@v2
       - uses: olafurpg/setup-scala@v10
```
jastice, line 17: could update this to v11 too

ckipp01: Done.

PR #187, opened 2021-06-28
Enforce scalafmt

ckipp01:
So this is a bit opinionated, but I noticed there was a `.scalafmt.conf` file, but it wasn't enforced at all. This also resulted in quite a few different indentation styles throughout the Scala sources. This pr adds in the sbt-scalafmt plugin, runs it with the existing configuration, and adds a check into CI to enforce formatting.

I don't have strong opinions on the type of formatting at all, just strong feelings on using it and being consistent. So if you want something switched, let me know. Or if you don't want this enabled at all, let me know and I can just remove the `.scalafmt.conf` file.

Note that I also added an extra commit to add in the `.git-blame-ignore-revs`. I've been using this lately in projects where I add in formatting since it combats the "now my git blame is useless" argument. Locally you can just:

```sh
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

And then your tooling that utilizes `git blame` will ignore the commit listed in the file.

Let me know your thoughts if this is all ok with you all.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,5 @@
+# It's a good idea to ignore the following commits from git blame.
+# You can read more about this here if you're unfamiliar:
+# https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame
+#
+ccb5152da9b44b5891858d9370ec944efcc5c49f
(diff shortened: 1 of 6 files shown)


adpi2 (approved):
Looks good to me.

I just learned something new with this git blame ignore file, really cool, thx!

ckipp01:
Since https://github.com/build-server-protocol/build-server-protocol/pull/194 was merged I updated this pr to just add in the checks and aliases, and then also updated the ignore refs file to point to the commit where steward formatted everything.

jastice (approved):
Works for me, but I'm not opinionated on formatting :)

PR #188, opened 2021-06-28
Update various deps.

ckipp01:
- jsoniter
- lsp4j
- ipcsocket
- scala-java8-compat
- scala-collection-compat
- sbt

This also removes sbt-coursier which really isn't being used or needed
seeing that we are using a modern sbt version.


jastice (approved):

build.sbt
```diff
@@ -103,10 +106,10 @@ lazy val tests = project
     publish / skip := true,
     libraryDependencies ++= List(
       "com.googlecode.java-diff-utils" % "diffutils" % "1.3.0",
-      "org.scala-lang.modules" %% "scala-java8-compat" % "0.9.0",
-      "org.scala-sbt.ipcsocket" % "ipcsocket" % "1.0.0",
+      "org.scala-lang.modules" %% "scala-java8-compat" % V.java8Compat,
+      "org.scala-sbt.ipcsocket" % "ipcsocket" % "1.4.0",
       "org.scalatest" %% "scalatest" % "3.0.9",
```
ckipp01, line 111: So this is a dep I didn't touch. I updated it to 3.0.9 when we cross published since this is the first version available for 2.13. However, 3.1.0 breaks compatability and rips out some off the generator stuff we need. So we can just pull a new dep in for that, but since there are some changes needed for that I was _curious_ if you've considered maybe switching to munit. It's almost a [drop in replacement](https://scalameta.org/munit/docs/scalatest.html) for scalatest, and imo a lot more user friendly. If you're satisfied with scalatest I can just go ahead and update it, but since it's the same amount of work to just switch to munit, I figured I'd ask 😄 .

ckipp01:
Also @jastice are you ok with me added `build-server-protocol/build-server-protocol` to the public Scala Steward repos? Then we can just get automated prs for these?

build.sbt
jastice: I have no strong opinion on this. Maybe someone else?

jastice:
> Also @jastice are you ok with me added `build-server-protocol/build-server-protocol` to the public Scala Steward repos? Then we can just get automated prs for these?

Sure why not!

PR #196, opened 2021-06-29
Add in `noOp` field to `CompileReport`.

### LINKED ISSUE #148: [Proposal] - A way to signify a no-op compilation
In [Metals](https://scalameta.org/metals/) we treat compilations a bit differently on whether or not we mark them as "no-ops". If a compilations is no-op it allows us to skip certain things in the editor UX such as refreshing the tree view. You can see a bit of this [logic here](https://github.com/scalameta/metals/blob/8f1c5db0e76ddbb95db06a55afc395ca6a965d13/metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala#L117-L148) if you're curious. We currently detect this when using Bloop by reaching into the `buid/taskStart` message to see if it contains a `Start no-op compilation`. You can see where this is marked in [Bloop here](https://github.com/scalacenter/bloop/blob/4768184cc8842d5d6773646ccda93210a89f1b05/frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala#L301-L331).

When we recently added support for sbt BSP, we obviously wanted to continue this distinction between _normal_ compilation and a _no-op_ compilation. However, that's hard to do when we're just relying on it being in a message. This lead me to wonder if it'd be a possibility to just bake this into the spec somehow as I can see how it'd be useful to have for various editors to determine what type of compilation task is taking place. I'm not sure where a logical place for this would actually be, but figured I'd throw the idea out there to see what others thought.

ckipp01:
This is meant to close #148 by adding in another field to the
`CompileReport` to signify to the client that the compilation was a
no-op. This is useful for some clients that trigger things based on
compliations, and they want to skip that trigger if the compilation was
a no-op. Currently this is accomplished by looking for keywords in the
message as is the case with Metals and Bloop.

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
@@ -436,6 +436,7 @@ class CompileReport {
   @NonNull Integer errors
   @NonNull Integer warnings
   Long time
+  Boolean noOp
   new(@NonNull BuildTargetIdentifier target, @NonNull Integer errors, @NonNull Integer warnings) {
     this.target = target
     this.errors = errors
--- a/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/CompileReport.java
+++ b/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/CompileReport.java
@@ -20,6 +20,8 @@ public class CompileReport {
   
   private Long time;
   
+  private Boolean noOp;
+  
   public CompileReport(@NonNull final BuildTargetIdentifier target, @NonNull final Integer errors, @NonNull final Integer warnings) {
     this.target = target;
     this.errors = errors;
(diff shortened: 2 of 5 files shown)


bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
```diff
@@ -436,6 +436,7 @@ class CompileReport {
   @NonNull Integer errors
   @NonNull Integer warnings
   Long time
+  Boolean noOp
```
ckipp01, line 439: I wasn't 100% sure to make the `@NonNull` or not. I'm not a huge fan of having `Option[Boolean]` in bsp4s, but at the same time I know that some servers won't set this, and I want to possibly avoid having a server set this as `false` giving the false impression that it's not a no-op when maybe it actually is. Is this ok like this, or do you think it'd be better modeled as `@NonNull`?

jastice (approved):
LGTM, for protocol changes let's get some more reviews.

adpi2 (approved):
I think it totally makes sense.

jastice:
I suppose this has mellowed enough to merge now

PR #258, opened 2021-12-14
[Build] `--release 8` for bsp4j

dos65:
This will help with `publishLocal`.
Otherwise you will have to switch jdk versions. 11 for xtend and 8 for publish


jastice:
Thanks for this!

PR #262, opened 2021-12-21
Update specification.md

Kludex:


jastice:
Good spot! Actually we use `SourcesParams` in the implementations, so I'll update it to that directly

PR #264, opened 2021-12-21
Update specification.md

Kludex:


jastice:
Thanks! Turns out this section needed more updates so I'll do that directly.

PR #269, opened 2021-12-28
Fix #205: Add operation to get output paths

### LINKED ISSUE #205: Exclude compilation artifacts from searchable files
IDEs usually provide a fast search feature, which allows you to search for some text in all the source files of your project. In the case of sbt projects, sbt generates a lot of small files in the `target/` directory. When we import a project via BSP in an IDE, these compilation artifacts (the content of `target/`) are included in the scope of a search, making it slow, and showing irrelevant results.

I think the developer experience would be improved if we introduced a concept of paths excluded from the actual sources of the project, that IDEs could leverage.

Currently, importing a project in IntelliJ via the sbt import feature does exclude the `target/` directories, but this is not the case when we import the same project via the BSP import feature (so, the sbt import does not suffer from this issue).

---

Here is a concrete example. I did `sbt new scala/scala3.g8`, and then imported the project (which uses sbt 1.5.5) in IntelliJ.

By default, IntelliJ imports the project via sbt, and the `target/` directory is excluded from the indexed files, as you can see here:

Screenshot from 2021-07-30 11-16-01

Then, if I re-import the project via BSP, the `target/` directory is not anymore excluded (it’s interesting to see that `project/target/` is excluded, though…):

Screenshot from 2021-07-30 11-20-23

As a consequence, when I search something, IntelliJ also searches in the `target/` directory:

Screenshot from 2021-07-30 11-22-07

Logs: [bsp-protocol-trace-2021-07-30T11-17-14.log](https://github.com/build-server-protocol/build-server-protocol/files/6906486/bsp-protocol-trace-2021-07-30T11-17-14.log)

povder:
This adds `outputDirectory` to BuildTarget to fix #205

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildServer.java
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildServer.java
@@ -37,6 +37,9 @@ public interface BuildServer {
     @JsonRequest("buildTarget/resources")
     CompletableFuture<ResourcesResult> buildTargetResources(ResourcesParams params);
 
+    @JsonRequest("buildTarget/outputPaths")
+    CompletableFuture<OutputPathsResult> buildTargetOutputPaths(OutputPathsParams params);
+
     @JsonRequest("buildTarget/compile")
     CompletableFuture<CompileResult> buildTargetCompile(CompileParams params);
 
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/OutputPathItemKind.java
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/OutputPathItemKind.java
@@ -1,0 +1,28 @@
+package ch.epfl.scala.bsp4j;
+
+import com.google.gson.annotations.JsonAdapter;
+import org.eclipse.lsp4j.jsonrpc.json.adapters.EnumTypeAdapter;
+
+@JsonAdapter(EnumTypeAdapter.Factory.class)
+public enum OutputPathItemKind {
+
+    FILE(1),
+    DIRECTORY(2);
+
+    private final int value;
+
+    OutputPathItemKind(int value) {
+        this.value = value;
+    }
+
+    public int getValue() {
+        return value;
+    }
+
+    public static OutputPathItemKind forValue(int value) {
+        OutputPathItemKind[] allValues = OutputPathItemKind.values();
+        if (value < 1 || value > allValues.length)
+            throw new IllegalArgumentException("Illegal enum value: " + value);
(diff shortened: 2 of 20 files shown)


jastice:
Thanks! I'm not sure this is the ideal way to address #205 rather than a more general "exclude" directory list for compilation artifacts that @olafurpg suggested. Also this has some overlap with `classDirectory` in [javacOptions](https://build-server-protocol.github.io/docs/extensions/java.html)

povder:
> I'm not sure this is the ideal way to address #205 rather than a more general "exclude" directory list for compilation artifacts that @olafurpg suggested.

I think you're right. I'll make a change to have a dedicated operation for obtaining "excluded" directories.

I'm not 100% sure about the naming though. "Exclude" says *what* clients should do with these directories (they should exclude them from indexing) but maybe BSP should instead say *what* these directories are and it's up to the IDEs to decide *what* to do with them.

jastice:
> I'm not 100% sure about the naming though. "Exclude" says _how_ clients should handle these directories (they should exclude them from indexing) but maybe BSP should instead say _what_ these directories are and it's up to the IDEs to decide _how_ to handle them.

Also a good point. So maybe just a general list of "outputs" (since it could be multiple). What do you think @olafurpg

olafurpg:
How about the name "generatedPaths"? It doesn't have to be a directory, it can also be a file. For example, Bazel auto-generates symbolic links. Otherwise I'm also OK with "outputPaths" or "outputs". I agree this should allow multiple paths.

jastice:
`outputs` sounds nice and general to me then, and it should be both directories and file paths. But I'm still fuzzy on the semantics - should this be only compile artifacts, or also generated sources? I think we would want the latter to be indexed.

ptrdom:
Seems like this PR has grown stale, do you need any help to proceed?

povder:
> Seems like this PR has grown stale, do you need any help to proceed?

Im sorry for letting this PR grow stale, I don't think I need any help before continuing the work. My plan is to work on this next week.

povder:
I added the `outputPaths` operation, it's ready for review.

jastice (approved):
Thank you!

povder:
Can I have somebody re-run the checks please? Thanks!

povder:
@jastice @tpasternak Can we have this released to maven central? I could then finalize my work in https://github.com/JetBrains/intellij-scala/pull/627 thx!

jastice:
`2.1.0-M3` is published :)

aishfenton:
Also opened an issue on the BSP spec, as it'd need to be solved there too. https://github.com/build-server-protocol/build-server-protocol/issues/446

Issue #249, opened 2021-11-30
Add `DebugSessionParamsDataKind` which will allow finer granulation of tests.


kpodsiad:
[Debug request](https://build-server-protocol.github.io/docs/specification.html#debug-request) has field `data` which content depends on `dataKind`. Currently, there is only one possibility to run/debug tests and it is `SCALA_TEST_SUITES` kind.
https://github.com/build-server-protocol/build-server-protocol/blob/2db6bd5093052058368c5390cfbbdcad4095bb4b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/DebugSessionParamsDataKind.java#L5

However, for this kind, the `data` field is an array of strings, for example `["a.b.myTestSuite, "a.b.c.AnotherTestSuite"]`, which is a very coarse granulation. A single test suite could consist of multiple test cases and it is impossible to represent a test run that consists of single or multiple (but not all) test cases from a particular test suite.

That's why I'd like to propose another `DebugSessionParamsDataKind` - (it's just a working name) `SCALA_TEST_CASES`. `data` for this kind can look like `TestCases[]`, where `TestCases` is defined as
```typescript
interface TestCases {
  suiteName: string,
  tests: string[]
}
```
Assuming that `["a.b.myTestSuite, "a.b.c.AnotherTestSuite"]` are test classes in project, such data structure will allow to represent:
- run for single test within single test suite- `[{suiteName: a.b.myTestSuite, tests: ["test 1"]}]`
- run for multiple test within single test suite - `[{suiteName: a.b.myTestSuite, tests: ["test 1", "test 3"]}]` // "test 2" is omitted
- run for multiple tests within multiple test suites - `[{suiteName: a.b.myTestSuite, tests: ["test 1"]}, {suiteName: a.b.c.AnotherTestSuite, tests: ["test 15"]}]`

`suiteName` together with the values of `tests` can be used to construct [test selector](https://github.com/sbt/test-interface/blob/master/src/main/java/sbt/testing/TestSelector.java) which can be further used to filter out unwanted test cases.

This proposition is based on my experience gained in https://github.com/scalameta/metals/pull/3277. Over the next days/weeks I'll try to do some more research or maybe even small proof of concept that will confirm if the proposed interface is sufficient, but I think it's a good time to open the issue and gather others' feedback about this.

adpi2:
That would be great.

[...]

PR #304, opened 2022-03-01
feat: add scala test selection request

kpodsiad:
implements #249

There are only two scala build servers that implement debugging: bloop and sbt (I checked also mill and bazel-bsp). If this new kind is implemented in both of them then it'll be safe to add the new kind without any capability.
Progress with implementation:
- [x] bloop (need changes)
- [x] sbt - https://github.com/scalacenter/scala-debug-adapter/pull/198
- [x] handling it in Metals - https://github.com/scalameta/metals/pull/3678

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -3,5 +3,4 @@
 -Xmx8G
 -XX:ReservedCodeCacheSize=1024m
 -XX:+TieredCompilation
--XX:+CMSClassUnloadingEnabled
 
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/DebugSessionParamsDataKind.java
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/DebugSessionParamsDataKind.java
@@ -3,5 +3,6 @@ package ch.epfl.scala.bsp4j;
 public final class DebugSessionParamsDataKind {
     public static final String SCALA_MAIN_CLASS = "scala-main-class";
     public static final String SCALA_TEST_SUITES = "scala-test-suites";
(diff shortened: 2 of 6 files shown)


jastice:
Wouldn't it be incompatible with prior versions of these build servers?

kpodsiad:
> Wouldn't it be incompatible with prior versions of these build servers?

The old request (with kind `scala-test-suites`) remains untouched and I'm adding another one, which is more universal. So if the old one remains, it shouldn't be incompatible.

I only wonder if a new capability is needed. Build client should be able to know if this request is supported. If yes, where should I put it? There is`BuildServerCapabilities`, but it isn't scala specific.

However, we may not add any capability because (I'm writing from the Metals perspective, I don't know if it valid for Intellij)
- for sbt, there is scala-debug-adapter plugin that adds this capability. Metals set particular version of this plugin 
- for bloop, Metals can check of version is greater than a particular one


After all, I think that adding a new capability may be better and more universal. What do you think?

tgodzik:
> Wouldn't it be incompatible with prior versions of these build servers?

I think since it's only Metals using it currently (I think Intellij doesn't use it?) it's fine as we are controlling the version of build servers and can send an appropriate request if we know a version doesn't support it. This can be done explicitly via capability, but I am not sure if it's really needed.

kpodsiad:
cc: @jastice @adpi2

adpi2 (approved):
LGTM. Also this is something I am looking forward to using in Metals.

jastice (commented):
LGTM, but maybe this mechanism is more general than Scala? Should we copy it to Java extension, or add it to JVM extension instead?

jastice (approved):

PR #312, opened 2022-03-31
Small nitpicks regarding JVM environment requests

kpodsiad:
Current documentation was outdated and it didn't reflect the actual shape of requests. Additionally, java classes lacked `originId` field - I added it with an additional constructor to not break compatibility.

For the reference this is what Scala classes looks like:
```scala
final case class JvmTestEnvironmentParams(
    targets: List[BuildTargetIdentifier],
    originId: Option[String]
)

final case class JvmEnvironmentItem(
    target: BuildTargetIdentifier,
    classpath: List[String],
    jvmOptions: List[String],
    workingDirectory: String,
    environmentVariables: Map[String, String]
)

final case class JvmTestEnvironmentResult(
    items: List[JvmEnvironmentItem]
)
```


adpi2 (approved):
LGTM!

jastice (approved):
Thanks!

Issue #319, opened 2022-04-08
Diagnostic out of sync with LSP


ckipp01:
Hello! I have a couple questions on the current format of `Diagnostic`. Currently in the BSP specification it says

> where Diagnostic is defined as it is in LSP.

This was probably true when BSP was first defined, but LSP has added a few more things to the LSP [`Diagnostic`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic). Just to name a few, `codeDescription`, `tags`, `data`. I'd like to know if the desire is to keep the BSP diagnostic in sync with the LSP diagnostic or to provide a minimal interpretation of it? I think it might be a good idea to mimic it, and if so I'd be happy to pr those changes in. If not, I can update the docs to notify users that it is different. If that is the case I'd like to propose that at least `codeDescription` be added in.

Another point is the `DiagnosticRelatedInformation`. I see that in the BSP implementation [here](https://github.com/build-server-protocol/build-server-protocol/blob/master/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/Diagnostic.java) that `relatedInformation` is just something that exists or doesn't, but is a single instance rather than a list of them like the LSP spec has them. I'm not sure if this was oversight or if there was a specific reason for this, but I was wondering if this could be aligned with the spec as well? I've been slowly working on [a plan](https://contributors.scala-lang.org/t/revisiting-dotty-diagnostics-for-tooling/5649) to expose more structured information in diagnostics, and part of that is utilizing `DiagnosticRelatedInformation`. The hope would be that these could be included right in [`xsbti.Problem`](https://github.com/sbt/sbt/discussions/6868) and then easily reported via BSP.  However, the use cases that I have for it have multiple different `relatedInformation`s that would need to be included with a diagnostic, which is currently impossible.

jastice:
> I'd like to know if the desire is to keep the BSP diagnostic in sync with the LSP diagnostic or to provide a minimal interpretation of it? I think it might be a good idea to mimic it, and if so I'd be happy to pr those changes in. If not, I can update the docs to notify users that it is different. If that is the case I'd like to propose that at least codeDescription be added in.

I think it would be fine to add in the additional data if we have any use case for it. Regardless, it seems like a good idea to explicitly specify the format of diagnostics rather than offhandedly refering to LSP spec.

[...]

PR #320, opened 2022-04-09
fix: align DiagnosticRelatedInformation with the LSP spec

ckipp01:
Previously this was a single DiagnosticRelatedInformation that was
possible to be attached to a diagnostic. This change aligns it with the
[LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic)
making this a `List[DiagnosticRelatedInformation]` instead.

Partly addresses #319

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/Protocol.xtend
@@ -250,7 +250,7 @@ class Diagnostic {
   String code
   String source
   @NonNull String message
-  DiagnosticRelatedInformation relatedInformation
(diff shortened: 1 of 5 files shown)


lukaszwawrzyk (approved):

jastice:
@adpi2 do you see any compatibility implications?

adpi2 (approved):
> @adpi2 do you see any compatibility implications?

I found no Scala open source tool that currently depends on `relatedInformation`. I checked Metals, Bloop, sbt and Mill.

Whether there are other tools out there that uses BSP, I don't know. If there are, chances are they don't use `relatedInformation` neither. So I am okay to break the compatibility here.

jastice (approved):

PR #363, opened 2022-09-15
feat: add `code` field to `Diagnostic`

lwronski:
This code field in Diagnostic is specified at https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic


tgodzik:
We specify in the documentation that `Diagnostic` is the same as the LSP one, so this is essentially a bugfix.

lwronski:
@jastice Could you take a look?  I fixed `sbt bps4j/xtend` and the tests are passed now.

tgodzik:
@jastice what do you think about this change?

adpi2 (approved):
LGTM

jastice:
I'm confused, the desciption says you're adding the `code`  field, but the PR adds a `data` field (also in spec) and a `dataKind` (which we use in BSP, but is not in `Diagnostic` spec). Can you clarify the intention?

lwronski:
> I'm confused, the desciption says you're adding the `code` field, but the PR adds a `data` field (also in spec) and a `dataKind` (which we use in BSP, but is not in `Diagnostic` spec). Can you clarify the intention?

Hi @jastice, I wanted to be consistent with the current implementation of bsp, where for every class that contains a `data` field, there is also the ` dataKind` field . For example in [DependencyModule.java](https://github.com/build-server-protocol/build-server-protocol/blob/db23ec3c210747347076878686b0082bb7b45b4b/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/DependencyModule.java#L18) or [BuildTarget.java](https://github.com/build-server-protocol/build-server-protocol/blob/db23ec3c210747347076878686b0082bb7b45b4b/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/BuildTarget.java#L32)

tgodzik:
I think that was a typo, we are adding the `data` field

jastice (approved):
thanks, I think that clears things up!

PR #367, opened 2022-09-22
Fix for `bsp4j/xtend`

lwronski:


tgodzik:
@jastice it seems this is another iteration of problems with `xtend`, could you take a look?

tpasternak:
Can we drop the whitespace changes?

lwronski:
> Can we drop the whitespace changes?

I can do it, but when I executed `sbt bsp4j/xtend`, sbt automatically generated whitespace changes for `bsp4j/src/main/xtend-gen` files.

PR #380, opened 2022-10-26
Adding mainClass parameter to JvmEnvironmentItem

sellophane:


bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/JvmEnvironmentItem.java
```diff
@@ -21,14 +21,18 @@ public class JvmEnvironmentItem {
   @NonNull
   private String workingDirectory;
 
+  @NonNull
```
jastice, line 24: Making this required break backwards compat, and since some implementations already rely on this we shouldn't without very good reason

jastice (changes requested):
I agree with Tomasz - a main class in a target is neither required, nor is there necessarily only one. Also requiring it at this point would require backward compat.

Also note, in the [Scala extension](https://github.com/build-server-protocol/build-server-protocol/blob/master/docs/extensions/scala.md) we have `buildTarget/scalaMainClasses` for this purpose, if necessary we can add such a request to the Java extension.

Also, do we really need this information at all? At least from IntelliJ perspective we can discover main classes on the client side.

sellophane:
Both mainClass and newly added javaArguments are required if we are planning on running a specific binary or test target from IJ knowing only its name.  Command line arguments are crucial for, among others, JUnit targets, because this it the way test package value is passed to JVM and if we want to run it locally (for example to be able to use IJ debugger) we need to know it.

sellophane:
So, after quick research it turned out that Scala CLI can have multiple main classes, so we need to have a collection here instead of a single instance. I made it nullable similarly to `ScalaMainClass. environmentVariables`

bsp4j/src/main/java/ch/epfl/scala/bsp4j/JvmExtension.xtend
```diff
@@ -31,13 +32,24 @@ class JvmTestEnvironmentParams {
   }
 }
 
+@JsonRpcData
+class JvmMainClass {
+  @NonNull @SerializedName("class") String className
+  @NonNull List<String> arguments
+    new(@NonNull String className, @NonNull List<String> arguments) {
+    this.className = className
+    this.arguments = arguments
```
tgodzik, line 41: Shouldn't the arguments be only defined when actually running the main? Why do we need it here?

jastice:
Sorry for the delay. I think this is good. Can we just add the changes to docs/specification?

sellophane:
> Sorry for the delay. I think this is good. Can we just add the changes to docs/specification?

Changed

jastice (approved):

ai-erickzhao:
Review

ai-erickzhao (commented):
Review After it's on Your Own Giving You access expect national security or something you can show as community can get satisfaction.

PR #386, opened 2022-11-09
Downgrade jsoniter to `2.13.5` to restore support for JDK 8

lwronski:
To unblock updating `bsp` in `bloop` I downgrade `jsoniter` to version `2.13.5` which support JDK 8.

Apply suggestion from https://github.com/scalacenter/bloop/pull/1845#issuecomment-1307234632.


plokhotnyuk:
Guys, please just help me in patching of https://github.com/sbt/sbt-multi-release-jar for working with JDK 11 and crossproject sbt plugins, and then I will release jars that works on JDK 8+ without performance loss on JDK 11+.

tgodzik:
> Guys, please just help me in patching of https://github.com/sbt/sbt-multi-release-jar for working with JDK 11 and crossproject sbt plugins, and then I will release jars that works on JDK 8+ without performance loss on JDK 11+.

I think it's not really an issue with  dropping JDK 8 as we should start dropping it and not waste time on maintaining everything for JDK 8, it's perfectly sensible that you dropped the support. We need to fix a couple of things first and then we would also drop JDK 8 support in Bloop.

lwronski:
> How about pinning this inside bloop?

The `jsoniter` from `bsp4s` is used transitively in bloop, so we don't have to pin `jsoniter` inside of bloop.

tgodzik:
> > How about pinning this inside bloop?
> 
> @tpasternak The `jsoniter` from `bsp4s` is used transitively in bloop, so we don't have to pin `jsoniter` inside of bloop.

We should do an override then, @tpasternak idea is that it will be easier for Bloop to just remove the override than doing the BSP release once more

lefou:
@tgodzik wrote:

> I think it's not really an issue with dropping JDK 8 as we should start dropping it and not waste time on maintaining everything for JDK 8, it's perfectly sensible that you dropped the support. We need to fix a couple of things first and then we would also drop JDK 8 support in Bloop.

Isn't the whole point of this project to support build tools to implement BSP? And isn't it in the nature of build tools that they need to play well on both edges of the version spectrum, newest JDK as well as oldest one that still gets security updates? Dropping support for JDK 8 means, tools like Mill, that provide BSP with the help of this very project, are forced to find an alternative solution to provide BSP for its users.

What is this huge benefit we get from dropping JDK 8 support, that adjusts such a cut?

plokhotnyuk:
Guys, I'm ready to cut a version with multi-release JAR that can run on JDK 8 and support all future updates. Please just help me hack the `build.sbt` to pack that kind of JAR for JVM sub-modules.

tgodzik:
> @tgodzik wrote:
> 
> > I think it's not really an issue with dropping JDK 8 as we should start dropping it and not waste time on maintaining everything for JDK 8, it's perfectly sensible that you dropped the support. We need to fix a couple of things first and then we would also drop JDK 8 support in Bloop.
> 
> Isn't the whole point of this project to support build tools to implement BSP? And isn't it in the nature of build tools that they need to play well on both edges of the version spectrum, newest JDK as well as oldest one that still gets security updates? Dropping support for JDK 8 means, tools like Mill, that provide BSP with the help of this very project, are forced to find an alternative solution to provide BSP for its users.
> 
> What is this huge benefit we get from dropping JDK 8 support, that adjusts such a cut?

Dropping JDK 8 would just be for Bloop. And we would still compile with the correct version, either via forked JVM for Java files, which already happens, or via the additional flags to make sure Scala code correctly compiles for a given JDK version. 

As for BSP, the Scala version is only used by Bloop and bsp4j works correctly with Java 8. Unless lsp4j library decides to drop the support at some point, which already happening to some other eclipse foundation libraries.

> Guys, I'm ready to cut a version with multi-release JAR that can run on JDK 8 and then support all future updates. Please just help me hack the build.sbt to pack that kind of JAR for the [jsoniter-scala-coreJVM](https://github.com/plokhotnyuk/jsoniter-scala/blob/e4bb1a28d672595b9ad4fed9680d40a34188847e/build.sbt#L144) sub-module.

Would love to help, but I think I know very little about sbt. Maybe the sbt channel on discord would be of help? Eugene should be very active there.

lwronski:
> How about pinning this inside bloop?

I have already overridden version of jsoniter in bloop - https://github.com/scalacenter/bloop/pull/1845.  So I will close this PR, because after discussion changes in `bsp` are not required.

Issue #147, opened 2020-11-03
Can search be enabled on the site?


ckipp01:
With Docusaurus you're able to enable search quite simply via `Algolia DocSearch`. You can find info about this `[here](https://docusaurus.io/docs/en/search)`. Then the snippet can just be `apiKey` can be entered into the `siteConfig.js`. I see it's already there actually (copied from Metals 👀), so it should be as simply as submitting the site to [here](https://docsearch.algolia.com/apply/) and your email, and uncommented out that chunk and getting search. I think this would be great as it's very useful for a doc site.

ckipp01:
Couple years later... wanted to revisit this. Would it be possible to get this enabled? As the site continues to grow, it'd be fantastic to have search.

jastice:
Finally got the key! Where should I put it? I've not really handled the docs before

ckipp01:
> Finally got the key! Where should I put it? I've not really handled the docs before

You should be able to uncomment and add it [here](https://github.com/build-server-protocol/build-server-protocol/blob/master/website/siteConfig.js#L19-L22)

PR #400, opened 2022-12-08
Update BSP website to use Docasaurus 2 

agluszak:
This will (among other things) cause the newly added extensions to appear in the sidebar.


jastice:
woohoo thanks! looks like test needs to be updated?

agluszak:
Fixes #147

PR #402, opened 2022-12-08
Fix pages deploy again

agluszak:


build.sbt
```diff
@@ -143,20 +143,6 @@ lazy val `bsp-testkit` = project
   .dependsOn(bsp4j)
   .enablePlugins(JavaAppPackaging)
 
-lazy val docs = project
```
ckipp01, line 146: Is this a mistake? This is using docusaurus, but it's using it through mdoc. By making this change you'll actually no longer be processing any of the mardown in `docs`. I don't believe you want tot do this, but instead should continue to rely on mdoc. If you want an example, [Metals](https://github.com/scalameta/metals) is also using V2 and being deployed with mdoc. You might want to look there for inspiration.

agluszak: Why do we want mdocs at all? I'm trying to get rid of this sbt-mdoc setup and simply use docusaurus for deploy

agluszak: Docusaurus takes care of generating html from md files, it works locally for me. I thought mdocs is used if you have scala code in the docs, but that's not the case here

ckipp01: The way the docs are set up for this site are reliant on mdoc. For example if you look in the `docs` directory, all of the code snippets are actually processed by mdoc when you see the code snippets like

```
```scala  mdoc
```

By removing your mdoc usage, you're breaking all of that. Again, it's not as simply as just removing mdoc and trying to deploy directly with docusaurus. This won't work.

agluszak: Okay, pardon, there's one place where we use mdoc: Java bindings documentation

ckipp01: Not only that, but the `VERSION` is also dynamically set this way.

agluszak: Hm, but anyway, mdoc is scala-specific, my students are working on rust bindings and and in general we want this to be language-agnostic, so using a scala-specific markdown processor is meh

ckipp01:
If I can kindly jump in here. Thanks a lot for working on this @agluszak. V2 in general seems to be a great improvement, however things are a bit broken at the moment. Here's what I noticed at quick glance

Screenshot 2022-12-08 at 20 02 53

- The sidebar is non-existent. This a regression from what we had before. Can this be brought back? The pr in https://github.com/build-server-protocol/build-server-protocol/pull/400 mentions a sidebar, but I don't believe it's working
- https://github.com/build-server-protocol/build-server-protocol/pull/400 also mentions that is fixes #147, but I don't see any search here. Was is supposed to be fixing that?
- Currently this is removing mdoc, I wouldn't recommend that. We'd be losing the typechecking on the docs that currently use this. I'd leave this for now.
- The base url has changed, breaking all previous links to the site

I'd recommend reverting the site back and first addressing the above points before moving forward.

build.sbt
tgodzik: > so using a scala-specific markdown processor is meh

Could you please elaborate? Why is this needed? I think using scala-specific markdown processor for scala examples is perfectly fine. As would rust specific one for rust. 

I would also maybe follow the traditional workflow with waiting for reviews, since this repository is important for a couple of tools and it would be better to make sure that nothing breaks.

ckipp01:
Superseded by https://github.com/build-server-protocol/build-server-protocol/pull/420.

PR #408, opened 2023-01-02
Adding tests and generated classes for JvmMainClass

sellophane:


jastice:
Thanks, can you also add the arbitrary/generators in `ch/epfl/scala/bsp/testkit/gen` please, so that roundtrip serialization is tested?

jastice:
looks like serialization roundtrip fails :)

testkit/src/main/scala/ch/epfl/scala/bsp/testkit/gen/Bsp4jGenerators.scala
```diff
@@ -644,13 +644,29 @@ trait Bsp4jGenerators {
...
+  } yield new JvmMainClass(className, arguments)
+
   lazy val genJvmEnvironmentItem: Gen[JvmEnvironmentItem] = for {
     target <- genBuildTargetIdentifier
     options <- arbitrary[String].list
     envVars <- genEnvironmentVariables.list
     classpath <- genFileUriString.list
     workdir <- genFileUriString
-  } yield new JvmEnvironmentItem(target, options, classpath, workdir, envVars.asScala.toMap.asJava)
+    mainClass <- genJvmMainClass
```
jastice, line 658: can add `.nullable` here to test the empty case

jastice (approved):
I think you can add the `.nullable`, otherwise LGTM!

jastice (approved):

PR #412, opened 2023-01-09
chore(ci): test on 8 and 17

ckipp01:
Since in theory we say we support 8, we should probably test against 8.
Since 17 is also being more commonly used, we should probably do that as well.
This also introduces windows to the matrix

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -6,13 +6,18 @@ on:
     branches: [master]
 jobs:
   generate-and-test:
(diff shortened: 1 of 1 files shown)


ckipp01:
Welp, nevermind I guess. It looks like the version of xtend we are using doesn't support 8.

jastice:
I guess we could do the matrix if we took xtend out of this part of the build? It's only there to double-check the code generation and is checked in, after all

PR #414, opened 2023-01-09
fix: ensure xtend was correctly ran

ckipp01:
It looks like previously extend wasn't ran meening that this was out of date a bit.


jastice (approved):

jastice:
on second look, why does xtend add that line to the constructor?

ckipp01:
> on second look, why does xtend add that line to the constructor?

To be honestly I'm not 100% sure. I just ran `xtend` and trusted it to do its job.

ckipp01:
It does match what is in https://github.com/build-server-protocol/build-server-protocol/blob/master/bsp4j/src/main/java/ch/epfl/scala/bsp4j/JvmExtension.xtend#L60.

ckipp01:
To be honestly I'm confused at how we can not have the latest generated code and all the tests still pass. We should probably have some sort of check to ensure that the generated code doesn't differ from what is checked in.

jastice:
In this case the change on generated code no-op, otherwise it would have been caught by the round-trip tests. But I agree it would be better to fully verify that generated code is up-to-date

PR #419, opened 2023-01-09
feat(ci): detect if there were file changes in CI when running `xtend`

ckipp01:
Alright you can see a working example of this in https://github.com/build-server-protocol/build-server-protocol/actions/runs/3875530094/jobs/6608192646. The idea here is that if you commit stuff without running `xtend` first you'll have changes in your Java files. When that happens, it will fail the build and we won't then test. This is meant to help avoid what happened in https://github.com/build-server-protocol/build-server-protocol/pull/414 where tests all pass and everything was green even though the new code wasn't generated.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
     branches: [master]
 jobs:
-  generate-and-test:
+  check-generation:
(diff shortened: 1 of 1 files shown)


jastice (approved):
cool, thanks!

PR #420, opened 2023-01-10
feat: update to docusaurus2

ckipp01:
This is a follow-up to the work started in
https://github.com/build-server-protocol/build-server-protocol/pull/400 updating docusaurus to v2. This should be a drop in replace and is very similar to the migration that we did in metals. This is also enabling search.

You can play around with the new version [here](https://bsp-ckipp01.vercel.app/).


tgodzik (approved):
LGTM from my side, thanks for doing this!

jastice (approved):
Cool, thanks for finishing this!

ckipp01:
There are some things on the dark mode that don't work well. I'll try to address those before merging.

jastice:
Can we merge now?

PR #424, opened 2023-01-11
Autogenerate sidebar items

agluszak:
Previously when we added a new extension we had to add it manually to the sidebar. But nobody was doing that, so the extensions didn't appear on the website. Now there's no need to change the sidebar manually.


agluszak:
image

agluszak:
I moved files around to enable autogeneration. And I also deleted `bsp.md` as it was only a redirection notice that would clutter the new autogenerated sidebar.

website/package.json
```diff
@@ -8,13 +8,12 @@
     "swizzle": "docusaurus swizzle",
     "docusaurus": "docusaurus"
   },
-  "devDependencies": {},
   "dependencies": {
     "@docusaurus/core": "2.2.0",
+    "@docusaurus/plugin-client-redirects": "2.2.0",
```
agluszak, line 13: idk why but yarn decided to reorder these items when I ran `yarn run`, let's leave it like that because sooner or later somebody else will probably run into the same thing

website/package.json
```diff
@@ -8,13 +8,12 @@
...
     "@docusaurus/preset-classic": "2.2.0",
     "clsx": "^1.1.1",
     "react": "^17.0.2",
-    "react-dom": "^17.0.2",
-    "@docusaurus/plugin-client-redirects": "2.2.0"
```
ckipp01, line 18: I thought this was needed for the redirects, but is that not the case?

ckipp01 (commented):
The few pages that are being moved, do you think we should add redirects in just in case?

website/sidebars.json
```diff
@@ -1,20 +1,39 @@
...
-  }
+  "docs": [
+    {
+      "type": "doc",
+      "label": "Specification",
+      "id": "specification"
+    },
+    {
+      "type": "category",
+      "label": "Concepts",
```
ckipp01, line 10: Also I don't think concepts is clear here. I think we should stick to overview. Plus then we don't need to change the existing urls.

agluszak: unfortunately to my knowledge it is impossible to setup autogenerate without providing a specific folder

agluszak: Regarding redirects - I don't think we need to be so paranoid about them. Even if for some reason someone decided to save a link to the FAQ, finding the new one is rather straightforward. We don't have that many pages anyway.

ckipp01: What I mean is can't this `label` just be `Overview`?

ckipp01: And regarding the redirects, there is a built in way to add them. I'd recommend just adding the redirects especially seeing you just changed probably the most visited page on the site.

website/docusaurus.config.js
```diff
@@ -40,8 +40,24 @@ module.exports = {
         ],
         "redirects": [
           {
-            "from": "/build-server-protocol/",
```
ckipp01, line 43: Silly `/` ha. I hope this works.

ckipp01 (approved):
Looks good! Really nice that we won't need to manually do this anymore. Thanks!

ckipp01:
Hey @agluszak just wanted to check back in. Was there anything else you were wanting to tackle with this, or is it good to merge?

PR #425, opened 2023-01-11
Run prettier in CI

agluszak:


.github/workflows/verify.yml
```diff
@@ -72,3 +74,19 @@ jobs:
           java-version: '11'
           cache: 'sbt'
       - run: sbt checkScalaFormat
+
+  website-prettier:
```
ckipp01, line 78: Not a blocker, but we could maybe add a check here to only run when there are any `*.yml`, `*.md`, or `*.js` files changed.

ckipp01 (approved):
🧹 up. Looks good to me!

ckipp01:
Small nit, if possible could you in the future create branches on your own fork instead of the main repo? By doing it your current way everyone following this repo will see it as activity on their main overview page. It's nice to not have this and typically a good idea to fork anyways.

agluszak:
> Small nit, if possible could you in the future create branches on your own fork instead of the main repo? By doing it your current way everyone following this repo will see it as activity on their main overview page. It's nice to not have this and typically a good idea to fork anyways.

I think you only get notifications if you set your follow settings to "everything", no? Unless you mean something else
image

ckipp01:
> I think you only get notifications if you set your follow settings to "everything", no? Unless you mean something else

Correct, but I like to turn this on for repos that I'd like to watch prs and issues on.

.github/workflows/verify.yml
agluszak: I'm not sure what you mean. If there are files of a type that prettier doesn't support, prettier doesn't run on them. And by definition no other file types will be places there, because why would they? Do you mean running this entire check conditionally? (I'm not sure if that's possible) Or do you mean running prettier only on files changed in that specific commit being checked?

ckipp01: yea I meant the entire thing, but no worries at all. Just ignore this.

ckipp01:
Hey @agluszak checking back in, was there anything left you were going to tackle with this, or is it good to be merged in?

Issue #410, opened 2023-01-09
[Proposal] Use the Smithy IDL as the source of truth for the BSP


Baccata:
## The context

Hi all. For those who don't know me, I'll introduce myself and give some context : my name is Olivier Mélois, I'm a principal engineer at Disney Streaming. I'm also an open-source maintainer and contributor, and am quite passionate about tooling. I have contributed some non-trivial stuff to mill and ammonite, and have some minor contributions to scala-cli and Intellij under my belt.

I maintain several open-source Scala libraries under the Disneystreaming flag :

* https://github.com/disneystreaming/weaver-test (a test framework)
* https://github.com/disneystreaming/smithy4s/  (a code-generation tool )

I also share a github organisation with my ex-colleague and friend, Anton Sviridov, which we called [neandertech](https://github.com/neandertech/), and where we maintain a small ecosystem related to the LSP, written in pure Scala : see [langoustine](https://github.com/neandertech/langoustine). 

I'm saying all this to give some context and show that I'm not (just) a crazy person coming to the repo with random ideas, but I like to think I put my money where my mouth is.

## The proposal

I'd like to propose for the source of truth to the BSP to be written in the [Smithy IDL](https://smithy.io/2.0/index.html).

### What is Smithy ?

Smithy is a protocol-agnostic IDL created by AWS. AWS uses Smithy as an input to code-generate all of the publicly available SDKs, as well as server-side bindings for the thousands of service they maintain in their platform. It has two syntaxes:

* the .smithy syntax, exemplified above
* a json syntax, which you can find [examples](https://github.com/aws/aws-sdk-js-v3/tree/main/codegen/sdk-codegen/aws-models) of in the AWS SDK repositories

The smithy metamodel (the building blocks used to construct smithy models) is really simple and intuitive, which makes it ideal as an input to code-generators.

Smithy comes with a number of fundamental building blocks to define data types and interfaces, in particular :

* structures
* unions
* collections (list/maps)
* a bunch of primitive types, including a `document` primitive representing raw json.
* operations (modelled as input/output/errors)
* services (collection of operations)

Additionally, Smithy decouples itself from protocol specific semantics via a powerful annotation mechanism that give meaning to the fundamental building blocks in specific protocols. For instance, it provides a bunch of [http specific](https://smithy.io/2.0/spec/http-bindings.html?highlight=http%20bindings) annotations.

### The Smithy tooling :

[...]

PR #426, opened 2023-01-12
Smithy as source of truth

Baccata:
See https://github.com/build-server-protocol/build-server-protocol/issues/410

At the time of writing this, the code-generation is bootstrapped. You can run it by calling `sbt bsp4j-gen/codegen` and some files should be created. Document generation is also bootstrapped in the docs module : running `sbt docs/mdoc` will dump a "specgen.md" in the generated website with documentation generated directly from the IDL. 

## TODO (for MVP) : 

- [x] Finish porting the spec to smithy (currently 100% of the markdown spec has been captured) 
- [x] Port things that are in bsp4j but not in the spec (c++ related things, for instance)
- [x] Render structures in Java
- [x] Render services in Java 
- [x] Render enumerations in Java 
- [x] Prep some typescript rendering to exemplify how the markdown documentation could be recreated 
- [x] Prep some documentation rendering (including typescript snippets) 
- [x] Wire documentation rendering to the existing docs module 
- [x] Wire smithy-codegen to xtend in the build 
- [x] Codify the open inheritance via some traits to facilitate the rendering of documentation
- [x] Check compatibility with the old bsp4j

### Noticed mistmaches between spec and bsp4j (WIP list) 

* inverseSources is declared as "textDocument/inverseSources" in the spec, as "buildTarget/inverseSources" in bsp4j


spec-traits/src/main/java/ch/epfl/smithy/jsonrpc/traits/EnumKindTrait.java
```diff
@@ -0,0 +1,87 @@
+package ch.epfl.smithy.jsonrpc.traits;
```
Baccata, line 1: the Java POJOs are there to facilitate the querying of information present in the smithy model : when assembling the model, the smithy-annotations automatically get deserialised to the registered POJOs. They're not 100% necessary, but they will be helpful to implement linters too later on.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,571 @@
+$version: "2"
```
Baccata, line 1: This file presents the most important part of the PR. It could be split into multiple smaller files without a problem, but I think it'll be up to the existing maintainers to decide how they want to split it.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,571 @@
...
+/// It is the server's responsibility to manage the lifetime of the diagnostics by using the appropriate value in the reset field.
+/// Clients generate new diagnostics by calling any BSP endpoint that triggers a buildTarget/compile, such as buildTarget/compile, buildTarget/test and buildTarget/run.
+@jsonNotification("build/publishDiagnostics")
+operation PublishDiagnostics {
+    input: PublishDiagnosticsParams
+}
+
+/// A resource identifier that is a valid URI according
+/// to rfc3986: * https://tools.ietf.org/html/rfc3986
+string URI
```
Baccata, line 97: You can define "type aliases" for primitive types, as such. It makes the spec easier to understand imho.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,571 @@
...
+    capabilities: BuildTargetCapabilities
+    /// Kind of data to expect in the `data` field. If this field is not set, the kind of data is not specified.
+    @required
+    dataKind: BuildTargetDataKind
+    /// Language-specific metadata about this target.
+    /// See ScalaBuildTarget as an example.
+    data: Json
+}
+
+@enumKind("open")
```
Baccata, line 171: there is no such things as "const" values in smithy. Instead, I've codified the distinction between **open enumerations** (which accept all values of the underlying type but expose a predefined set), and **closed enumerations** (which only accept the values that are defined), as a custom smithy trait that I called `enumKind`, defined alongside the jsonRPC protocol.

spec/src/main/resources/META-INF/smithy/bsp/tags.smithy
```diff
@@ -0,0 +1,36 @@
+$version: "2"
+
+namespace bsp
+
+// Applying the tags a-posteriori to avoid hurting the readability of the spec
+apply URI @tags(["basic"])
```
Baccata, line 6: as you can see in this file, `traits` can be applied a-posteriori : they are a little bit like Scala typeclasses in that regard. 

Because I wanted to avoid bloating the main spec file with tags, I've started tagging the shapes here in a separate file. 

The `@tags` trait could be later used to group shapes together to control how they appear in documentation.

spec/src/main/resources/META-INF/smithy/jsonrpc/jsonrpc.smithy
```diff
@@ -0,0 +1,43 @@
+$version: "2"
+
+namespace jsonrpc
```
Baccata, line 3: this files defines the various traits that give meaning to the BSP specification in the context of jsonRPC.

codegen/src/main/scala/bsp/codegen/Lines.scala
```diff
@@ -0,0 +1,59 @@
+package bsp.codegen
+
+import cats.kernel.Monoid
+
+trait Lines { self =>
```
Baccata, line 5: This is a construct that's convenient to manipulate for indentation and stuff. We use a similar thing in smithy4s, though more complex.

codegen/src/main/scala/bsp/codegen/IR.scala
```diff
@@ -0,0 +1,61 @@
+package bsp.codegen
+
+import software.amazon.smithy.model.shapes.ShapeId
```
Baccata, line 3: This file represents the "intermediate representation" that acts as a middleman between the smithy-model and the renderer. It contains the cherry-picked, sufficient information for re-creating the bsp4j code.

build.sbt
```diff
@@ -165,3 +260,90 @@ addCommandAlias(
   "checkScalaFormat",
   "scalafmtCheckAll ; scalafmtSbtCheck"
 )
+
+// Bootstrapping task that wires the build to the
+// codegen module's main method
+def runCodegen(config: Configuration) = Def.task {
```
Baccata, line 266: This is the build task that wires invokes the code-generator via shell-out. I'm not sure the caching was necessary in the scope of this project, but, I've just copied/pasted/adapted what we do in smithy4s.

codegen/src/main/scala/bsp/codegen/ModelLoader.scala
```diff
@@ -0,0 +1,10 @@
+package bsp.codegen
+
+import software.amazon.smithy.model.Model
+
+object ModelLoader {
+
+  // By virtue of this module dependong on the `spec` module, the models can be found on the classpath.
```
jastice, line 7: typo: "dependong"

Baccata:
I'm happy to report that the full markdown specification has been ported. I did it manually so there may be a couple mistakes in there.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,1380 @@
...
+/// Like the language server protocol, the initialize request is sent as the first request from the client to the server.
+/// If the server receives a request or notification before the initialize request it should act as follows:
+///
+/// * For a request the response should be an error with code: -32002. The message can be picked by the server.
+/// * Notifications should be dropped, except for the exit notification. This will allow the exit of a server without an initialize request.
+///
+/// Until the server has responded to the initialize request with an InitializeBuildResult, the client must not send any additional
+/// requests or notifications to the server.
+@jsonRequest("build/initialize")
+operation InitializeBuild {
```
ckipp01, line 54: I'm not really sure how to phrase the question, so bear with me. Looking at the docs and such I see the `operation`s can also have `errors`. In this scenario where the server receives a request or notification before the initialize and the response is supposed to be a `-32002`, is that able to be captured at all in the spec here? I know it doesn't belong in `InitializeBuild.errors`, but is there anything that can capture that?

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,1380 @@
...
+    input: WorkspaceBuildTargetsParams
+    output: WorkspaceBuildTargetsResult
+}
+
+/// The `reload` request is sent from the client to instruct the build server to reload
+/// the build configuration. This request should be supported by build tools that keep
+/// their state in memory. If the `reload` request returns with an error, it's expected
+/// that other requests respond with the previously known "good" state.
+@jsonRequest("workspace/reload")
+operation WorkspaceReload {
```
ckipp01, line 124: Another question about `errors` in `operation`s. So in here I know it says:

> code and message in case an error happens during reload. For example, when the build configuration is invalid.

So are these supposed to actually be encoded in here or no? Like we know that this can error and that it's supposed to have a code and a message. I supposed if we'd do that we'd need to create a `GenericBspError` structure or something like that _if we wanted to do it_?

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,1380 @@
...
+/// for example be used by a language server before `textDocument/rename` to ensure
+/// that all workspace sources typecheck correctly and are up-to-date.
+@jsonRequest("buildTarget/compile")
+operation BuildTargetCompile {
+    input: CompileParams
+    output: CompileResult
+}
+
+@jsonRequest("buildTarget/test")
+operation BuildTargetTest {
```
ckipp01, line 238: Should this be included in the `BuildServer.operations`?

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
ckipp01: Ah nice, I was just about to come and comment and this and saw your not. Do we actually want this to be `open`? What happens if we do actually get a tag for example that we don't know about. Shouldn't it be closed forcing one of these?

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,1380 @@
...
+}
+
+string Identifier
+
+list Identifiers {
+    member: Identifier
+}
+
+/// Included in notifications of tasks or requests to signal the completion state.
+@enumKind("open")
```
ckipp01, line 461: Same comment as above, I'm inclined to say this should be `closed`. What are you thoughts about keeping it `open`?

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,1380 @@
...
+    dependencies: BuildTargetIdentifiers
+    /// The capabilities of this build target.
+    @required
+    capabilities: BuildTargetCapabilities
+    /// Kind of data to expect in the `data` field. If this field is not set, the kind of data is not specified.
+    @required
+    dataKind: BuildTargetDataKind
+    /// Language-specific metadata about this target.
+    /// See ScalaBuildTarget as an example.
+    data: Json
```
ckipp01, line 377: I mean... who am I to complain.. this is literally what the spec says, but I wish this was more constrained. Like we know it will be a `SomethingBuildTarget` _right_? I can already see myself wanting tighter contraints than what the spec calls for, but I imagine that might be dangerous if we loosen up in the future and allow more things in `data`. But I wish `data` here could just be an `ADT` of like `BuildTargetData` where we had `ScalaBuildTarget`, etc as members.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,1380 @@
...
+
+structure OnBuildShowMessageParams with [MessageParams] {
+
+}
+
+structure LogMessageParams with [MessageParams] {
+
+}
+
+@enumKind("open")
```
ckipp01, line 643: Should this be open?

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,1380 @@
...
+    tags: DiagnosticTags
+    /// An array of related diagnostic information, e.g. when symbol-names within
+    /// a scope collide all definitions can be marked via this property.
+    relatedInformation: DiagnosticRelatedInformationList
+    /// A data entry field that is preserved between a `textDocument/publishDiagnostics` notification
+    // and a `textDocument/codeAction` request.
+    data: Json
+}
+
+structure Position {
```
ckipp01, line 703: Maybe we do somewhere else, but it might be a good idea to mention if these are 0 or 1 based.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,1380 @@
...
+    /// The text document's URI.
+    @required
+    uri: URI
+}
+
+list Diagnostics {
+    member: Diagnostic
+}
+
+@enumKind("open")
```
ckipp01, line 734: Same old closed comment. No idea but I think this one might want to be closed as well.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,1380 @@
...
+    data: Json
+}
+
+list BuildTargetEvents {
+    member: BuildTargetEvent
+}
+
+/// The `BuildTargetEventKind` information can be used by clients to trigger
+/// reindexing or update the user interface with the new information.
+@enumKind("open")
```
ckipp01, line 822: Closed?

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,1380 @@
...
+
+  /// Timestamp of when the event started in milliseconds since Epoch.
+  eventTime: EventTime,
+
+  /// Message describing the task.
+  message: String
+
+  /// Kind of data to expect in the `data` field. If this field is not set, the kind of data is not specified.
+  /// Kind names for specific tasks like compile, test, etc are specified in the protocol.
+  dataKind: TaskDataKind
```
ckipp01, line 1064: Out of the scope of this pr, but just a general thought. I wish we could restrict the type of `data` here based of `TaskDataKind`.

ckipp01: This idea of a value restricting what `data` can be seems pretty common, but we just don't capture that in a structured way which is a bummer.

ckipp01 (commented):
Really great job on this @Baccata, super impressive. I'm slowly going through it all and I just got through the `.smithy` files and have a handful of questions, so I figured I'd pause here. Another one not in an individual comment, but since I've seen very little smithy (and you commented slightly on this), what's best practice for breaking stuff up? Do you keep all `service` and `operation`s in one file and then `structure`s in another? Probably _it depends_ is the answer here.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
Baccata: It'd certainly be possible to define a `jsonRPCError` trait, use it to annotate structures and list those structures in operations. Then, code generators could potentially generate those structures as throwable an apply some special handling in the routing logic to apply the correct error code when such throwable are caught. 

However, this is way beyond the scope of this PR, the goal of which is to try and capture the information present in the existing specification as opposed to improving / filling the gaps of the existing specification.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
Baccata: > I supposed if we'd do that we'd need to create a GenericBspError structure or something like that if we wanted to do it? 

The way it'd look like would probably be : 

``` 
@jsonRPCError(-32130) // or whatever 
structure ReloadError {
    message: String 
    whateverElse: Something 
}
```

and the data contained by this ReloadError` shape would be filling the `data` of the error payload.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
Baccata: Good catch, fixed

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
Baccata: > But I wish data here could just be an ADT of like BuildTargetData where we had ScalaBuildTarget, etc as members. 

Right, so this is a very open question and there are several ways to approach it, but in the context of BSP, I believe no approach is perfect and will be subject to trade-offs, in particular in light of the following problems : 

#### Problem 1 : the BSP is an open protocol 

Doing what you ask means unavoidably creating a stronger bias towards Scala and Java, because suddenly you have stronger constraints for those things than you have for the rest. 

So when you say "Like we know it will be a SomethingBuildTarget right ?", I think the answer is "no, you don't know, because you've made BSP an open protocol as opposed to a closed one". 

It's basically like the difference between sealed traits and open-inheritance. You're asking the question implying that you want sealed-traits in a context where the decision was very much to leave the data unsealed so that the protocol could be re-used in other contexts than Scala/Java (unless I'm massively mis-understanding the scope of BSP) 

#### Problem 2 : the existing data types as they are defined in the BSP are ... not great 

The data field and discriminator are inlined with the rest of the information (like capabilities) and whatnot. This breaks the inductive nature of encoders/decoders : if we do what you want the way you want it, instead of code-generators being able to simply defer to macro-based derivation provided out of the box by Json libraries, the code-generators suddenly have to contain logic to ensure that the data is encoded correctly. 

So an alternative would be that the whole `BuildTarget` would become an ADT, where each of its member would abide by a shared interface and where only the `data` field would change. This is certainly possible to encode in Smithy, but it means the source of truth becomes less intuitive. 

#### Observation : 

There are a few knobs to play with : you can find ways of encoding things such that the spec would still be naive and give a pretty good idea of what the end json looks like, whilst tuning what the code-generators do to add more constraints when you use bsp4s (or bsp4j). 

To give you some intuition of what I mean : in Scala, you can easily decouple the data definition from the Json encoders/decoders. For instance, you can have `case class FooBar(foo: String, bar: String)` that could be encoded in json as `fooValue@barValue`. 

When people read `case class Foobar(...)`, what they expect the json will look like is 

```json 
{
  "foo" : "fooValue", 
  "bar" : "barValue"  
} 
```

Now say you need to retrofit some json value that looks like `fooValue@barValue` in an IDL. What do you do ? Do you say : 

1. "this is a string and it must look like this", or 
2. "it's a structure, but I'm gonna annotate it this special way to let everyone know it ought to be encoded this way" 

And that's the fundamental trade-off : if you go with 1, you make the spec precise but easier to implement. If you go with 2, you make the spec more precise but harder to implement. 
 
#### Conclusion : 

You need to be **extremely precise about what you're asking for**, and understand the trade-offs. I purposefully adhere to the spec in a somewhat oblivious manner because it's already gonna be a leap for the BSP maintainers to get used to the smithy language and metamodel and going with a fancy encoding for these things would make the leap even longer.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
Baccata: > What happens if we do actually get a tag for example that we don't know about. Shouldn't it be closed forcing one of these? 

Well you tell me : do you want to seal your protocol or leave it open for other usecases than Scala and SBT. I think the latter. If you get a tag you don't know about, the party that doesn't know about it would probably fail one way or another. 

Like, what if you have a Kotlin BSP server ? The `"kotlin"` value should be valid as per the protocol, doesn't mean that Metals should be able to handle "kotlin".

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
Baccata: The fact that is open is accurate as per the current specification. 

I cannot comment as to whether this is correct or not 🤷 From an open-protocol standpoint, which BSP aims at being, as far as I follow, it makes sense to accept any int value. For instance, you could have a BSP client and a BSP server that would have an understanding of `4` representing "DELAYED". If you go from open to close, it means you're preventing that. 

**However**, you can have bespoke code-generators that would generate open-enumerations as closed ones 🤷‍♂️. In Scala, the difference is basically between :  

```scala 
enum StatusCode {
   case Ok 
   case Error
   case Cancelled 
} 
``` 

and 

```scala 
enum StatusCode {
   case Ok 
   case Error
   case Cancelled
   case Other(value: Int) 
} 
```

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
Baccata: Your prerogative, but this is faithful to the current spec. If you wanted a closed enumeration, the typescript should have looked like : 

``` 
type MessageType = 1 | 2 | 3 | 4
```

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
Baccata: Addressed.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
Baccata: Ditto https://github.com/build-server-protocol/build-server-protocol/pull/426/files/c43f4e3150a006b1a511ce59a695418f8360005e#r1070611189

spec/src/main/resources/META-INF/smithy/jsonrpc/jsonrpc.smithy
Baccata: > how does that work if they want to make an official one 

There is no global trait index. In the same way that there is no official Json library in Scala. It might be a good idea to change the namespace to `bsp.jsonrpc` though, as, if this project could probably claim the `bsp` namespace in a global index, it'd be much less the case for `jsonrpc`. 

If/when something else defines its own jsonrpc semantics, an approach that I've used that kinda works is to pre-process the models to turn one set of semantics to another (when possible), to gain access to whatever that "something else" does.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
Baccata: See my response [there](https://github.com/build-server-protocol/build-server-protocol/pull/426#discussion_r1070608894)

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
Baccata: Here are some more details that can fuel the discussion : 

* Smithy4s uses smithy-unions to represent sealed-coproducts, and provides 3 ways of encoding unions out of the box : https://disneystreaming.github.io/smithy4s/docs/codegen/unions
* AWS says that in the context of their protocols, unions are always encoded as tagged unions (read the link above to understand the difference). Arguably tagged unions are the best encoding for unions, in that this encoding can be parsed with no backtracking, and is the most flexible in terms in what you can put in the alternatives. So AWS is not wrong in how it tries to be super-constraining with unions.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
ckipp01: Gotcha, makes sense. Thanks!

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
ckipp01: Gotcha, makes sense.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
ckipp01: Yea, go with what we currently have then. To be honest I'm not 100% sure what is there now was actually mean to be that way, open, or if it was just an oversight. Either way, go with what exists now.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
ckipp01: Alright, makes sense 👍🏼

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
ckipp01: Alright, thanks the extra context here is super helpful.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
jastice: Yep, this is very much intentionally open. The protocol definition here gives a few known use cases, but any implementor is free to define new values on client or server side, and either side should be able to handle unknown values

codegen/src/main/scala/bsp/codegen/TypescriptSpecRenderer.scala
```diff
@@ -0,0 +1,121 @@
...
+import software.amazon.smithy.model.shapes.ShapeId
+
+import dsl._
+import bsp.codegen.EnumType.IntEnum
+import bsp.codegen.EnumType.StringEnum
+import bsp.codegen.JsonRPCMethodType.Notification
+import bsp.codegen.JsonRPCMethodType.Request
+import bsp.codegen.Hint.Documentation
+
+class TypescriptSpecRenderer(basepkg: String) {
```
jastice, line 16: Rendering TS may come in handy anyway, but I wonder if we should just migrate the main spec to Smithy, so there's less of a disconnect between source of truth and documentation. Is there a straightforward way of doing that?

Baccata: You mean advertise the smithy file as documentation ? Or generate markdown with smithy snippets in it ?

Either way, I think the best course of action is to defer that decision. This POC should, ihmo, strive to recreate something reasonably close to the current state. 

Smithy is not as well known as typescript by any mean, and offering documentation with typescript snippets kinda make sense, in particular if those snippets are automatically rendered as opposed to handcrafted.

If later on you want to make it more dynamic and render the snippets in the markdown differently, it'd be possible. I think this PR proves that code generation from smithy is not difficult, whatever the target language is.

Baccata: Smithy provides a renderer out of the box but the atomic unit of rendering you get is the namespace : the function that renders individual shape is private. You could ask Aws whether they'd be keen to make it public but I have no idea whether they'd agree to it.

Regarding this renderer, I'm writing it to use for markdown but you could obviously change the code in slight ways to render an actual typescript library

docs/specgen.md
```diff
@@ -0,0 +1,94 @@
...
+The base protocol is identical to the language server base protocol. See
+<https://microsoft.github.io/language-server-protocol/specification> for more
+details.
+
+Like the language server protocol, the build server protocol defines a set of
+JSON-RPC request, response and notification messages which are exchanged using
+the base protocol.
+
+```scala mdoc:passthrough
+bsp.codegen.Codegen.printDocs()
```
Baccata, line 93: This wires the code-generation of docs to the markdown. The structuring of the spec changes a little to avoid doing convoluted shoehorning in this unit of work, but the heavylifting is mostly done.

Baccata:
@jastice @ckipp01 so at this point, the heavy-lifting is mostly done. 

I do think I need to change how open-inheritance is codified in order to facilitate the rendering of the markdown documentation, as I'm not really satisfied with the current state in this PR. 

I also think that at this point, enough has been achieved that it'd be profitable to have a video call with the stakeholders to talk about : 

1. what is in this PR so far 
2. the various way that the open-inheritance could be codified in smithy, and the induced tradeoffs 
3. the concerns that the stakeholders may have with this proposal 
4. action points, etc

Without wanting to put much pressure on you all, I'd appreciate if a date for a sync could be set promptly (even if the date in question is a few weeks from now), as I've invested a fair amount of time to ensure that all technical locks are lifted.

codegen/src/main/scala/bsp/codegen/MarkdownRenderer.scala
```diff
@@ -0,0 +1,116 @@
...
+    val renderer = new MarkdownRenderer(state)
+    renderer.render(tree).get.mkString(System.lineSeparator())
+  }
+}
+
+class MarkdownRenderer private (state: MSet[ShapeId]) {
+
+  import dsl._
+
+  def render(tree: DocTree): Lines = {
```
ckipp01, line 22: Just an idea, it might be nice to make a `Renderer` trait with a single abstract `render` if that's the only thing that's really exposed. For no other reason than to cement the pattern of having a renderer that exposes a render just to make it clear in the future when we want to create another one, we just create a new renderer and implement `render`.

codegen/src/main/scala/bsp/codegen/ToLines.scala
```diff
@@ -0,0 +1,17 @@
...
+
+object ToLines {
+
+  def apply[A](value: A)(implicit ev: ToLines[A]): List[String] = ev.lines(value)
+
+  implicit val stringToLines: ToLines[String] = (a: String) => List(a)
+
+  implicit def flattenedLines[A: ToLines]: ToLines[List[A]] = (a: List[A]) => a.flatMap(ToLines(_))
+
+  implicit def linesToLInes: ToLines[Lines] = (a: Lines) => a.get
```
ckipp01, line 15:
```suggestion
  implicit def linesToLines: ToLines[Lines] = (a: Lines) => a.get
```
Not sure if this was on purpose or not

codegen/src/main/scala/bsp/codegen/Lines.scala
ckipp01: Just for clarity, can you add some scaladoc explaining this, especially since there are other things related to it like the `ToLines` and the implicits there. I want to solidify for future people why we have this.

ckipp01 (commented):
Just a few other random things I noticed. Again, great job on this @baccata. Looking forward to the overview!

ckipp01:
Alrighty, jumping back in here after our meeting last week. Again, thanks for taking the time @Baccata and going through all of this with everyone. I'll outline what I understood leaving the meeting below, and please do correct me if I misunderstood something or if you have more input.

## Yea or Nay

If I understood the general atmosphere in the meeting, we all seemed to be on the same page about moving forward with this. I guess this is your last chance to speak up if you don't think this is the direction we should head and why.

## Next Steps

Baring someone speaking up on the above point, and also taking into account the checklist that @Baccata has on the pr description, here's a few things I think we'd want to tackle before getting this merged in and moving forward:

_Taken from the original list_

- [ ] Port things that are in bsp4j but not in the spec (c++ related things, for instance)
- [ ] Wire smithy-codegen to xtend in the build*

* Regarding this one, I think instead of actually wiring it straight into the build at this point, I'd propose we do this in multiple steps
- [ ] After everything is confirmed ported, ensure that we can fully generate bsp4j
- [ ] Document steps to generate
- [ ] Add in another roundtrip test like we have but swap out the current bsp4j with the newly generated one just to make sure it passes

Once the above is done I'd propose that we actually merge in to not get this PR any larger. After that is done then we'd do the following:

- [ ] Get rid of the old bsp4j and replace it with the generated one
- [ ] If possible, get rid of xtend altogether, and just generate bsp4j directly

This isn't taking into account the doc generation or other generation, but the main focus of the above would be bsp4j, and then we can also tackle the doc generation along with it or after, whatever you recommend @Baccata.

## Who does this

@Baccata you expressed making sure you had someone from this side also investing some time and helping. If no one else that was in the meeting wants to dig in, I'm happy to spend some time with you getting this across the board.

Let me know what you all think, or even just give a 👍🏼 if you're all good with this.

Baccata:
I'm happy with this plan !

jastice:
Thank you @ckipp01 for keeping this moving! I probably don't have much time myself for doing much work on this so your involvement is really appreciated. But let me know if I can help out.

agluszak:
> Port things that are in bsp4j but not in the spec (c++ related things, for instance)
c++ is and was in the spec, but until #424 the extension was not being shown on the website. This extension, however, is not used by any clients afaik. It was one of the first extensions ever added and was, let's say, experimental...

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -0,0 +1,1332 @@
...
+/// * a single configuration in a single project that contains both Java and Scala sources maps to one BuildTarget.
+/// * a project with crossScalaVersions 2.11 and 2.12 containing main and test configuration in each will have 4 build targets.
+/// * a Scala 2.11 and 2.12 cross-built project for Scala.js and the JVM with main and test configurations will have 8 build targets.
+/// * Pants: a pants target corresponds one-to-one with a BuildTarget
+/// * Bazel: a bazel target corresponds one-to-one with a BuildTarget
+///
+/// The general idea is that the BuildTarget data structure should contain only information that is fast or cheap to compute.
+structure BuildTarget {
+    /// The target’s unique identifier
+    @required
```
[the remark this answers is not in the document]
Baccata: Unfortunately not.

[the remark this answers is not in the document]
Baccata: Or rather, theoretically speaking you could come up with something but you'd be going against the idiom, which I don't think is a good thing.

build.sbt
```diff
@@ -165,3 +256,104 @@ addCommandAlias(
...
+                .withRunJVMOptions(Vector("-cp", classpath))
+                .withOutputStrategy(
+                  CustomOutput(outputStream)
+                )
+
+              val exitCode = Fork.java(options, mc +: args)
+
+              if (exitCode != 0) {
+                s.log.error(outputStream.toString())
+                throw new RuntimeException(s"Codegen failed with exit code $exitCode")
```
Baccata, line 344: 👍

Baccata:
@agluszak I won't comment on the semantic changes you made to the bsp.smithy files as you know the domain better, but I think it's a good decision to put language specific things in their own files 👍 

I have just one nitpick on the change of modelling of the `Field` case class, but really it's a detail. I think the rest looks good. 

Is there a particular piece of logic you wanted me to focus my review on ?

agluszak:
Current problems with the docs:
- [ ] ensure the correct order of the "common shapes" (BuildTarget first etc.)
- [ ] render simple shapes (URI, Identifier) with type aliases (currently they show as strings)
- [ ] finish extension rendering

lefou (approved):

PR #466, opened 2023-03-25
refactor: switch from Algolia to a local search

ckipp01:
This replaces the current Algolia search (which doesn't work) with
`easyops-cn/docusaurus-search-local`, which seems to work pretty well.
We use this in
[Metals](https://github.com/scalameta/metals/pull/5066/files) as well,
which is what gave me the idea to switch it here seeing that it's
currently not working.

_You can see an example of this working here_

https://user-images.githubusercontent.com/13974112/227724242-1779cae8-acbf-476f-8257-2bbc4e56b907.mov


adpi2 (approved):
LGTM

PR #516, opened 2023-05-09
Add skip debugger configuration

adpi2:


adpi2:
Finally I decided to use the `noDebug` field in the `LaunchRequest` of DAP instead.

PR #526, opened 2023-05-19
Type rendering fixes

### LINKED ISSUE #521: The type flaws in the protocol
1. Prefer use primitive types rather than boxed type

Please refer: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#general-types
I suggest to replace 'String' with 'string', and also for Boolean or anything mentioned in the link...

2. Use integer rather than Integer, or define Integer
This is a really small flaw, more like typo.
The integer type defined in LSP is 'integer', refer https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#baseTypes 

3. `Long` is not defined
This is bit worse than `Integer`, because this type is not defined!

4. Some types are not defined
Like this
``` ts
export interface TestParams {
  // ...

  /** Kind of data to expect in the `data` field. If this field is not set, the kind of data is not specified. */
  dataKind?: String;

  /** Language-specific metadata about for this test execution.
   * See ScalaTestParams as an example. */ // This is also strange, because I could not find ScalaTestParams in search of the website
  data?: TestParamsData; // HERE, the type "TestParamsData;" is lacked
}
export namespace TestParamsDataKind { // And this type is not used at all.
  /** `data` field must contain a ScalaTestParams object. */
  export const ScalaTest = "scala-test";
}
```

And this happens for nearly all `data?: XXXX`

Seems we want to express that the implementation could give the data whatever they like.

Then I suggest to replace the design with:
``` ts
export interface TestParams {
  // ...

  /** Language-specific metadata about for this test execution. */
  data?: TestParamsData;
}

export interface TestParamsData{
  kind: TestParamsDataKind;
  [key: string]: any;
}

export type TestParamsDataKind = string;

export namespace TestParamsDataKind {
  /** `data` field must contain a ScalaTestParams object. */
  export const ScalaTest = "scala-test";
}
```

And I suggest more strict constraint to the name of TestParamsDataKind, like `${language_name}-test`.

5. Wrong use of namespace in typescript

Well, although I understand the intention is like "includes but not limited". And it's much more important to understand the design rather than flaws like this, it's not correct to write something like
``` ts
namespace Foo{
export const foo1 = "123"
}
interface Bar{
  foo: Foo; // ts will complains about this in fact.
}
```

You need to define Foo firstly:

``` ts
type Foo = string;
namespace Foo{
export const foo1 = "123"
}
interface Bar{
  foo: Foo; // Good now
}
```

Just like what I suggested in the fourth point  :)

agluszak:
Fixes #521 

- [x] Use unboxed types
- [x] Define Integer and Long
- [x] Fix wrong use of namespace in typescript
- [ ] Render FooData types as aliases to any


jastice (approved):

codegen/src/main/scala/bsp/codegen/ToLines.scala
```diff
@@ -12,6 +12,8 @@ object ToLines {
 
   implicit def flattenedLines[A: ToLines]: ToLines[List[A]] = (a: List[A]) => a.flatMap(ToLines(_))
 
-  implicit def linesToLInes: ToLines[Lines] = (a: Lines) => a.get
+  implicit def linesToLines: ToLines[Lines] = (a: Lines) => a.get
+
+  implicit def optionToLines[A: ToLines]: ToLines[Option[A]] = (a: Option[A]) => ToLines(a.toList)
```
ckipp01, line 17: Maybe I'm missing the usage, but isn't this unused?

codegen/src/main/scala/bsp/codegen/docs/MarkdownRenderer.scala
```diff
@@ -25,69 +32,88 @@ class MarkdownRenderer private (tree: DocTree, visited: MSet[ShapeId]) {
...
           newline,
-          tree.commonShapes.foldMap(renderNode),
+          tree.commonShapeIds.foldMap(renderStructureNode),
           newline
         )
 
-    val services = tree.services.foldMap(renderNode)
-    commonShapes ++ services
+    // Server comes first because it's bigger.
+    val services = tree.services.sortBy(-_.operations.size).foldMap(renderServiceNode)
```
ckipp01, line 45: I find `-_.` a bit funky here. Wouldn't
```suggestion
    val services = tree.services.sortBy(_.operations.size).reverse.foldMap(renderServiceNode)
```
Be the same and maybe easier to read when glancing over this?

codegen/src/main/scala/bsp/codegen/docs/MarkdownRenderer.scala
```diff
@@ -25,69 +32,88 @@ class MarkdownRenderer private (tree: DocTree, visited: MSet[ShapeId]) {
...
-            )
-          } ++ lines(
-            members.foldMap(renderNode)
-          )
+        lines(
+          header,
+          newline,
+          operations
+            .map(tree.operations)
+            .foldMap(renderOperationNode)
```
ckipp01, line 115: Feel free to ignore this since you didn't add it here, but I'm not sure how others feel about using deps transitively without explicitly declaring them in the build. I see we're using `foldMap` here from cats, but we never define a cats dependency. Again, no strong feelings on this but I typically like to declare something if we're explicitly using it.

ckipp01 (approved):
Nothing blocking, just a couple small comments. But other than that, LGTM

codegen/src/main/scala/bsp/codegen/docs/MarkdownRenderer.scala
```diff
@@ -25,69 +32,88 @@ class MarkdownRenderer private (tree: DocTree, visited: MSet[ShapeId]) {
...
+      documentation(definition.hints),
+      newline,
+      tsBlock(definition),
+      newline
+    )
+  }
+
+  def renderOperationNode(node: OperationDocNode): Lines = {
+    node match {
+      case OperationDocNode(operation, input, output) =>
```
adpi2, line 85: I suggest to avoid pattern matching on a single pattern.

```suggestion
    val OperationDocNode(operation, input, output) = node
```

codegen/src/main/scala/bsp/codegen/docs/MarkdownRenderer.scala
```diff
@@ -25,69 +32,88 @@ class MarkdownRenderer private (tree: DocTree, visited: MSet[ShapeId]) {
...
+          input.foldMap(renderStructureNode),
+          output.foldMap(renderStructureNode)
         )
+    }
+
+  }
+
+  def renderServiceNode(node: ServiceDocNode): Lines = {
+    node match {
       case ServiceDocNode(shapeId, operations) =>
```
adpi2, line 104: Same here:
```suggestion
     val ServiceDocNode(shapeId, operations) = node
```

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -9,6 +9,18 @@ use jsonrpc#jsonNotification
...
+integer Integer
+
+/// A long is a 64-bit signed integer ranging from -2^63 to (2^63)-1 (inclusive).
+long Long
+
+/// A float is a single precision IEEE-754 floating point number.
+float Float
+
+/// A double is a double precision IEEE-754 floating point number.
+double Double
```
adpi2, line 22: It seems we don't use `Double` and `Float` anywhere. Will we ever need them? Should we remove them? Or at least hide them?

adpi2 (approved):
Some minor comments, but other than that LGTM

codegen/src/main/scala/bsp/codegen/ToLines.scala
agluszak: yeah, that's a leftover from the previous implementation

PR #550, opened 2023-06-19
fix: various prettier fixes

ckipp01:
This PR makes a couple small changes:

- adds prettier as a dev dependency to `website/package.json` to more easily be able to run it locally
- changes CI to not try and commit but instead just to check the formatting.
- also just inlines the formatting changes for prettier to the normal verify.yml file


jastice:
Previously the formatter check caused issues when updating the readme: It would only accept the specific `prettier` style, and you'd have to run `prettier` to get it to pass. The tool requires a separate installation rather than being part of the regular build, so that makes simple changes to readme needlessly tedious. So imho we'd want one of

- only check for syntactic validity of markdown
- automatically reformat and commit
- use a formatter that can be installed and called from sbt

ckipp01:
> Previously the formatter check caused issues when updating the readme: It would only accept the specific prettier style, and you'd have to run prettier to get it to pass. 

Sure, but by definition this is what a format check does.

> The tool requires a separate installation rather than being part of the regular build, so that makes simple changes to readme needlessly tedious.

Sure, but the way it's set up now, there is no way to locally run prettier unless you have prettier installed locally. If we are saying we want prettier to format our project, then there _must_ be a way to manually do it if you'd like with the provided tools. Since we have a `package.json`, then it should be set up to simply do a `npm run format` to actually format the project. Having something enforced and setup in CI with no way to do the same locally will put us in a position like we're in now where we have CI failing due to formatting, and no way locally to fix it unless you globally install a tool. We shouldn't do this.

> only check for syntactic validity of markdown

If this is all you want, then we probably shouldn't be using prettier.

> automatically reformat and commit

I personally really dislike auto reformat and commit. But if others don't mind, then I guess ok, but we should still have a way to do it locally as well.

> use a formatter that can be installed and called from sbt

I don't really think this is realistic. Prettier is coming from the JS ecosystem, managing it with npm seems natural and fine since we're already using it in our project.

jastice:
> Sure, but by definition this is what a format check does.

A format could be permissive, at least in principle

> Sure, but the way it's set up now, there is no way to locally run prettier unless you have prettier installed locally. If we are saying we want prettier to format our project, then there must be a way to manually do it if you'd like with the provided tools.

Exactly, that's my gripe :) I guess this PR somewhat fixes it?

> If this is all you want, then we probably shouldn't be using prettier.

IMHO that would be my preference, I'm not too fond of strict format enforcement, but I won't argue the point unless it gets in the way of doing simple fixes to the docs (as it did previously)

> I personally really dislike auto reformat and commit. But if others don't mind, then I guess ok, but we should still have a way to do it locally as well.

agree

> I don't really think this is realistic. Prettier is coming from the JS ecosystem, managing it with npm seems natural and fine since we're already using it in our project.

I guess my complaint was having 2 build tools on this project and having to install npm effectively just to update the readme

ckipp01:
> > Sure, but the way it's set up now, there is no way to locally run prettier unless you have prettier installed locally. If we are saying we want prettier to format our project, then there must be a way to manually do it if you'd like with the provided tools.
> 
> Exactly, that's my gripe :) I guess this PR somewhat fixes it?

Indeed, yes.

> I guess my complaint was having 2 build tools on this project and having to install npm effectively just to update the readme

But we already use npm, locally if you want to test the website, you have to use npm. That's the main reason from my why it doesn't seem like a big deal.

jastice:
Fair enough, though I'm not super happy about having to run a formatter to fix a readme

jastice (approved):

PR #552, opened 2023-06-22
Refactor approval tests

agluszak:
Move approved files generated by tests to the `resources` folder


ckipp01 (approved):
👍🏼 just to make the reviews easier do you mind including a brief description in the future of an overview of what changes are being made and the rational behind them. It'll also make the git history read better in the future if we ever revisit any of these changes.

Issue #506, opened 2023-04-22
WIP: Minor inconsistencies discovered during the smithy migration


agluszak:
- [ ] Sometimes we use a map for passing environment variables (https://build-server-protocol.github.io/docs/extensions/jvm#test-environment-request), sometimes we use a list (https://build-server-protocol.github.io/docs/extensions/scala#scala-specific-data-kinds-for-debugging)

agluszak:
Related: #562

Issue #480, opened 2023-04-08
What does the version number mean?


agluszak:
Currently the architecture of this repository resembles a monorepo, as it consists of many independent components:
- the BSP spec itself
- bsp4j library
- bsp4s library
- website generator

In January we "released" version [v2.1.0-M4](https://github.com/build-server-protocol/build-server-protocol/releases/tag/v2.1.0-M4).
But that was a release of what? The protocol itself didn't change. Some dependencies were bumped. The biggest change was the new website generator. 

I propose that we disentangle versioning/releasing schemes of each component.

I guess that it initially made sense to keep the protocol and the support library versions in lockstep. But now it's starting to become a burden - introducing a minor change to one of libraries shouldn't necessitate bumping the version of the protocol itself. And vice versa - rephrasing a sentence in the spec doesn't necessarily make the libraries incompatible with the new version of the protocol. This problem is going to grow bigger and bigger as more support libraries are added.

Changes to the website generator shouldn't affect the version number at all.

lefou:
Version 2.1.0-M4 changed the protocol a bit, as it added some `classes` fields to the `JvmEnvironmentItem`, IIRC.

+1 for separating the version number from the support libraries. Also, I'd like to see separate changelogs for protocol and libraries.

If versions get split, it would be nice if the support libraries get some API which exposes the supported BSP protocol version.

adpi2:
> Changes to the website generator shouldn't affect the version number at all.

That I agree.

But for splitting the spec and the libraries I am not quite sure it is good idea:
- we don't want the libraries and the spec to diverge: when we change something in the spec, we want to see that change in the lib
- it is rare to change something in a lib without changing the spec: the libs are more or less just the code generated from the spec.
- having the same version is convenient for the user
- having three separate things to release is more cumbersome

ckipp01:
I think traditionally this was also done because its... easier to maintain. 

> But now it's starting to become a burden

Could you expand a bit on what the burden is?

I also sort of agree with @adpi2 that every time the actual spec changes there needs to be those changes reflected as well in the implementations, so I'm not convinced that we gain a lot by doing this. The only benefit I can see would be having bug fixes in the implementations wouldn't bump that spec when it's not needed. That might be nice, but again, if there isn't an actual burden I'm not sure it's worth it.

lefou:
[...]

PR #555, opened 2023-06-23
Replace hand-written bsp4j and bsp4s with generated versions

agluszak:
Some things are going to break because bsp4j, bsp4s and the typescript definition of the protocol sometimes didn't agree one with another. In other cases the libraries were internally inconsistent because of copy-pasting.

What's going to break:

In bsp4j:
- `onConnectWithServer` is removed - it wasn't a proper json-rpc notification anyway - probably a leftover from some older version? Also it wasn't defined in the spec at all.
- `MessageType.INFORMATION` is renamed to `MessageType.INFO`. The spec clearly says `INFO`: https://build-server-protocol.github.io/docs/specification#messagetype
- fields in `JvmBuildTarget` structure were optional in the protocol and bsp4s, so in the corresponding bsp4j class they were moved from constructor parameters to setters. We can either change (the protocol AND bsp4s) or bsp4j. I chose the latter option. But I don't know if these fields being optional make any sense.
- analogous changes in the cpp extension, but it's not used by anyone anyway
- `DebugSessionParams` [had an obligatory `dataKind`](https://github.com/build-server-protocol/build-server-protocol/pull/555/files#diff-048020527ac3494dbf4c3a9f95197b9266ce8bfa5f1bf7d66f40e1a51636d916L16) which was illegal according to the spec

In bsp4s:
- `CleanCacheResult` no longer has a `message` field. It wasn't defined in the spec.
- all bsp enums now use `value` for getting the inner value (previously some used id, some used code and some used value)

In spec:
- Made BuildTargetCapabilities' fields optional (as null == false in this case). This isn't a breaking change to the spec, but it will break some code in bsp4j (although it will save some other code from breaking, depending on which constructor it uses). See https://github.com/build-server-protocol/build-server-protocol/pull/555#pullrequestreview-1507435869

Non-breaking changes:
- some classes should have had a `dataKind`, but they didn't. Now they do.
- `${language}Extension`.xtend files are broken into separate classes for each structure
- in bsp4s order of definitions is changed - now they are sorted alphabetically
- whitespace changes, I tried to minimize them as much as I could, but again, they were inconsistent in the first place
- hashcode/equals definitions changed (again, they were inconsistent in the first place) and other minor things. For example see https://github.com/build-server-protocol/build-server-protocol/pull/555/files#diff-3a02edaa2405332587baa34149132faeb8012baf0121ff07a1092df100332592 and https://github.com/build-server-protocol/build-server-protocol/pull/555/files#diff-19d83b3030b2fdb7887c55d73380153fd476265395bb70851a1890fbed419b4a
- in `JvmEnvironmentItem` there was this [useless identity assignment ](https://github.com/build-server-protocol/build-server-protocol/pull/555/files#diff-4dbed8893e4ba0e02ed3d25d37e4fcfa26c7079d6daaaf20d0da01f0d53334f0L35)

All of the changes above come from the fact that now everything is generated from the smithy model in a consistent way. Hand-written implementations were inconsistent in many places. I didn't introduce any special cases/hardcode anything. But as you can see, all of the breaking changes are very minor/simple.

In bsp4s there's a constant `Bsp4s.ProtocolVersion` and in bsp4j `Bsp4j.PROTOCOL_VERSION` with the supported protocol version.

To test this you'll need to do the following

1. Run `sbt generate`
3. Rest `sbt test` to show that all the normal tests are still passing, etc


jastice (commented):
In this PR we have a lot of breaking method name changes, such as `buildInitialize` -> `initializeBuild`. From separate discussion, I understand this is for consistency of naming. I also understand we couldn't keep the current naming of everything and have it be both consistent and api/binary compatible. However I'd prefer to keep incompatible changes to a minimum. What do others think?

agluszak:
I don't think there are that many users of our libraries, so introducing these (minor tbh - that's a few renames and moving setters to constructor parameters or vice versa in a few classes) breaking changes will not cause a leftpad incident ;)

ckipp01:
I'll try to review this early next week when I'm back, but since this does include some breaking changes I think we should also ping

- @tgodzik 
- @lefou 

Just so they are aware ASAP and can also provide some input on the breakage.

bsp4j/src/main/java/ch/epfl/scala/bsp4j/RunParamsDataKind.java
```diff
@@ -1,5 +1,4 @@
 package ch.epfl.scala.bsp4j;
 
 public class RunParamsDataKind {
-    public static final String SCALA_MAIN_CLASS = "scala-main-class";
```
agluszak, line 4: Fixed in #561

agluszak:
> Could we put the renames into a separate commit, so that they are easier to review?

not really, but I did my best to list all the breaking changes in the PR description

jastice:
> buildInitialize is renamed to initializeBuild and buildShutdown to shutdownBuild. Their params were named "InitializeBuildParams" and "ShutdownBuildParams", so to keep things consistent we either have to rename the method names or the params. All other params are named "${methodName}Params}". BTW, I think we should remove the "build" part of it altogether, because these methods are not related in any way to building. They are related to the lifetime of the server.

They are named based on the json-rpc method name, like `build/initialize`. Other methods such as `buildTarget/resources` still retain the name `buildTargetResources` in bsp4j. So I think we should try to be consistent in the name transformation and maybe rename the parameters instead.

> in every extension buildTarget${name}Options method is renamed to ${name}Options (the base protocol doesn't use the buildTarget prefix in method names)
It does, as in above example

> In ScalaWorkspaceEdit the list of changes is optional. That's how it was defined in the protocol ¯_(ツ)_/¯ (I agree, it doesn't make any sense, but again, we either break the protocol or the library)
The addition is still new, maybe we can fix it?

> in bsp4s order of definitions is changed (from one seemingly random ordering to another seemingly random ordering; I can sort them alphabetically for example)
I guess alphabetical makes sense for minimal conflicts in future additions

agluszak:
> It does, as in above example
Yes, of course you are right. I'll fix it then.

> I guess alphabetical makes sense for minimal conflicts in future additions
I'll fix it in the next commit.

> This too.

bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildTargetTag.java
```diff
@@ -2,8 +2,8 @@
 
 public class BuildTargetTag {
     public static final String LIBRARY = "library";
-    public static final String TEST = "test";
     public static final String APPLICATION = "application";
+    public static final String TEST = "test";
```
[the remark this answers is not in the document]
agluszak: let's

agluszak:
I reverted the renames and opened #562

bsp4j/src/main/java/ch/epfl/scala/bsp4j/MessageType.java
```diff
@@ -8,7 +8,7 @@ public enum MessageType {
 
     ERROR(1),
     WARNING(2),
-    INFORMATION(3),
+    INFO(3),
```
jastice, line 11: technically should be INFO, though as long as it maps to the right value it shouldn't matter. But we can change it if we still have other breakages

bsp4j/src/main/java/ch/epfl/scala/bsp4j/DiagnosticTag.java
```diff
@@ -0,0 +1,6 @@
+package ch.epfl.scala.bsp4j;
+
+public class DiagnosticTag {
+    public static final int DEPRECATED = 2;
```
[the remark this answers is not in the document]
agluszak: let's sort only string ones then :)

bsp4j/src/main/java/ch/epfl/scala/bsp4j/MessageType.java
agluszak: we do still have other breakages and I'd prefer not to write a special case for INFO -> INFORMATION transformation

agluszak:
Question to SBT wizards: what should I do to have `codegen` and `xtend` tasks run automatically in sequence after each `compile` task?

jastice:
> Question to SBT wizards: what should I do to have codegen and xtend tasks run automatically in sequence after each compile task?

You *could* do that, but I don't think you'd want to. `compile` supports incremental compilation, the code gen afaik doesn't.

bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/ScalaTestClassesItem.java
```diff
@@ -11,21 +11,14 @@ public class ScalaTestClassesItem {
...
 
-  private String framework;
-
   public ScalaTestClassesItem(@NonNull final BuildTargetIdentifier target, @NonNull final List<String> classes) {
     this.target = target;
     this.classes = classes;
-    this.framework = null;
-  }
-
-  public ScalaTestClassesItem(@NonNull final BuildTargetIdentifier target, @NonNull final List<String> classes, final String framework) {
```
jastice, line 25: Looks like this just removes a convenience constructor. Is this beneficial?

agluszak:
There's also #506

bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/ScalaTestClassesItem.java
agluszak: It was the only class with this kind of constructor. Again, I think we should either generate such constructors for every class or for none of them. Otherwise we'd have to introduce a special case in the generator just for this one class.

tests/src/test/scala/tests/SerializationSuite.scala
```diff
@@ -44,7 +44,7 @@ class SerializationSuite extends AnyFunSuite {
     assert(bsp4jValue.getTaskId.getParents.asScala == bsp4sValue.taskId.parents.get)
     assert(bsp4jValue.getEventTime == bsp4sValue.eventTime.get)
     assert(bsp4jValue.getDataKind == bsp4sValue.dataKind.get)
-    assert(bsp4jValue.getStatus.getValue == bsp4sValue.status.code)
+    assert(bsp4jValue.getStatus.getValue == bsp4sValue.status.id)
```
jastice, line 47: Where does the change from `code` to `id` come from?

agluszak:
> You could do that, but I don't think you'd want to.
Because the entire codebase for bsp4j and bsp4s are now generated, there's never a situation where you'd want to run a `compile` without running `codegen` and `xtend` beforehand.

bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/ScalaTestClassesItem.java
jastice: At least one more class has it: `ScalaWorkspaceEdit`
I'm mostly agnostic on this, but I wouldn't mind having a "full" convenience constructor for every such class. Would reduce some boilerplate setter calls in tests.

tests/src/test/scala/tests/SerializationSuite.scala
agluszak: It can be called value or code if you prefer

jastice:
> Because the entire codebase for bsp4j and bsp4s are now generated, there's never a situation where you'd want to run a compile without running codegen and xtend beforehand.

I suggest creating a dedicated task for generating and compiling then to not mess with expected behavior of tasks too much. Something like 

```sbt
val buildGeneratedCode := {
  codegen.value
  xtend.value
  compile.value
}
```

My sbt is rusty, might not fully work like this

tests/src/test/scala/tests/SerializationSuite.scala
agluszak: But it will have to be the same for all enums

bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/ScalaTestClassesItem.java
agluszak: But if we include a "full" constructor it will quickly become a problem once we add a new field to an existing struct. Because adding a new field will cause the generator to change the "full" constructor and therefore introduce a breaking change

tests/src/test/scala/tests/SerializationSuite.scala
```diff
@@ -113,33 +112,12 @@ class SerializationSuite extends AnyFunSuite {
...
     assert(bsp4jDiagnostic1.getMessage == bsp4sValue.diagnostics.head.message)
     assert(
-      bsp4jDiagnostic1.getSeverity.getValue == bsp4sValue.diagnostics.head.severity.get.id
+      bsp4jDiagnostic1.getSeverity.getValue == bsp4sValue.diagnostics.head.severity.get.value
     )
     assert(bsp4sValueDecoded.diagnostics.head.data == diagnostic1.data)
     assert(bsp4sValueDecoded == bsp4sValue)
   }
 
-  test("BuildTargetCapabilities - backward compatible canDebug") {
```
jastice, line 122: what happened to this?

bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/ScalaTestClassesItem.java
agluszak: If you are worried about boilerplate in case of code we maintain - in the Kotlin library we'll have data classes. And if someone else wants to get rid of boilerplate in their java code they can contribute a generator for a new library which will use java 17 records ;)

jastice: > But if we include a "full" constructor it will quickly become a problem once we add a new field to an existing struct

Good argument, so better to remove them fully.

tests/src/test/scala/tests/SerializationSuite.scala
agluszak: BuildTargetCapabilities in bsp4s had a `canDebug: Boolean = false` field. It was the only field with a default value, so I didn't write a generator for default values. `canDebug` became an obligatory field in the spec some time ago

agluszak: I have an idea now: we can make all fields of `BuildTargetCapabilities` optional in the spec, as null amounts to the same as false in this case.

jastice: sounds good to me

jastice (approved):
So I think we can live with some minor breaking changes on generated API level, so from my side it's good to go! I'd like to get feedback from the other tool maintainers though.

lefou:
I guess I can live with API breakages if that means, we are stricter following the protocol. In Mill 0.11, we dynamically load bsp4j into the server and do not expose it to our users.

But breaking the API means, we need to bump the version of the implementation(s) in an appropriate way (semver), which also means, we are no longer in sync with the protocol version. I recall #480 and think we should decouple the protocol version from the implementation version(s) and introduce some API that returns the implemented protocol version. This also allows to publish maintenance releases without any protocol changes, which is IMHO a good thing. Especially, if some transitive dependencies can and should be bumped to fix some (security) issues.

agluszak:
> Generally agree, though it might lead to some confusion. Might dual versions make sense here? Something like `bsp4j-bsp_2.7-3.1`?

If we include Scala cross-compilation  in bsp4s we'll have *triple* versions :D

lefou:
> > But breaking the API means, we need to bump the version of the implementation(s) in an appropriate way (semver), which also means, we are no longer in sync with the protocol version.
> 
> Generally agree, though it might lead to some confusion. Might dual versions make sense here? Something like `bsp4j-bsp_2.7-3.1`?

I don't see that much potential for confusion. But to avoid them from the start, we could add a table to the readme, explaning which implementation version implements which protocol version. 

Dual versions come with their own issues, e.g. we might end up with multiple incompatible versions in the classpath, due to transitive dependencies and distinct artifact names. Also, if we ever choose dual versioning, we always need a prefix in the version part (e.g. `_bsp2.1-M5`), as the non-prefixed version suffix is already taken by Scala.

bsp4s/src/main/scala/ch/epfl/scala/bsp/endpoints/Endpoints.scala
```diff
@@ -3,63 +3,45 @@ package endpoints
...
-
-  // Scala specific endpoints
-  object scalacOptions
-      extends Endpoint[ScalacOptionsParams, ScalacOptionsResult]("buildTarget/scalacOptions")
-  @deprecated("jvm test environment should be used instead")
-  object scalaTestClasses
-      extends Endpoint[ScalaTestClassesParams, ScalaTestClassesResult](
-        "buildTarget/scalaTestClasses"
-      )
-  @deprecated("jvm run environment should be used instead")
```
ckipp01, line 58: It looks like we lose these `@deprecated` annotations down below right? Do we want to even have these methods anymore, of if we are breaking things should we just remove them?

ckipp01 (commented):
At a high level this looks great. Nice job! 

I think we spoke about this a bit in the past about what we check in and what we don't, but I think we might want to re-visit this now. All the xtend files sort of become this intermediate format that we generate only to use to generate java code. Do we actually want to check these in? Or does it make sense now to stop check them in and only check in the Java code that is generated? What benefit is there to checking in the xtend files?

I also think since we're generating bsp4s now as well we should addd a check in the `verify.yml` just like we do for java and ensure that the generated scalafiles aren't different then what is checked in. Related, currently in CI we only ever use `xtend` to generate stuff when we do this check. We should adjust this to do the full generation from Smity to xtend, to Java/Scala to really ensure what we are doing is what is checked in.

jastice:
> Or does it make sense now to stop check them in and only check in the Java code that is generated? What benefit is there to checking in the xtend files?

I agree. We could even consider not checking in anything generated, but I think there's value in having the code browsable.

agluszak:
I removed xtend files from the repo, java files generated from xtend will now live in the same package as those generated directly

bsp4s/src/main/scala/ch/epfl/scala/bsp/endpoints/Endpoints.scala
agluszak: fixed

agluszak:
Marked 2 methods in the scala extension as deprecated and implemented a renderer for deprecations

agluszak:
> I also think since we're generating bsp4s now as well we should addd a check in the verify.yml just like we do for java and ensure that the generated scalafiles aren't different then what is checked in. Related, currently in CI we only ever use xtend to generate stuff when we do this check. We should adjust this to do the full generation from Smity to xtend, to Java/Scala to really ensure what we are doing is what is checked in.
 
 Fixed

agluszak:
I think all remaining issues have been addressed :) @ckipp01 @jastice @tgodzik @lefou

agluszak:
Unfortunately our property tests are flaky... [This current run](https://github.com/build-server-protocol/build-server-protocol/actions/runs/5466195994/jobs/9950801348?pr=555) failed because a URI generator generated an incorrect URI (not related to my changes in any way). [Some time ago](https://github.com/build-server-protocol/build-server-protocol/actions/runs/5424552233/jobs/9864101249) a run failed because the property generator didn't stop after 6 hours. This happened to me locally a few times as well. Again, it's not related in any way to this PR.

lefou:
I might have overlooked it, as this PR is quite large. Is there an API now, to get the implemented protocol version? Or did you decide to make version-specific artifacts?

jastice:
> I might have overlooked it, as this PR is quite large. Is there an API now, to get the implemented protocol version? Or did you decide to make version-specific artifacts?

I don't think we've decided anything :)
Adding a version API seems like the easier thing though. Any opinions?

agluszak:
But there's always been a request for getting the BSP version: in the [build/initialize](https://build-server-protocol.github.io/docs/specification#initializebuild-request) request first the client says the version of BSP it uses and then the server replies with its BSP version

jastice:
> But there's always been a request for getting the BSP version: in the build/initialize request first the client says the version of BSP it uses and then the server replies with its BSP version

The issue here though is knowing which BSP version the library supports, as opposed to the server. It should be somehow clear or discernible from the programmer's perspective. It could be as simple as a hard coded string that can be reall easily navigated to.

lefou:
> > But there's always been a request for getting the BSP version: in the build/initialize request first the client says the version of BSP it uses and then the server replies with its BSP version
> 
> The issue here though is knowing which BSP version the library supports, as opposed to the server. It should be somehow clear or discernible from the programmer's perspective. It could be as simple as a hard coded string that can be reall easily navigated to.

Exactly. I'd like to use that library hard-coded version to use in the BSP Server implementation to answer the question, which BSP version does this server speak.

agluszak:
Ahh, I see now. Okay, so let me implement it.

agluszak:
Implemented @lefou @jastice.

In bsp4s there's a constant `Bsp4s.ProtocolVersion` and in bsp4j `Bsp4j.PROTOCOL_VERSION`

lefou:
> Implemented @lefou @jastice.
> 
> In bsp4s there's a constant `Bsp4s.ProtocolVersion` and in bsp4j `Bsp4j.PROTOCOL_VERSION`

Great, thanks!

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Bsp4j.java
```diff
@@ -0,0 +1,5 @@
+package ch.epfl.scala.bsp4j;
+
+public class Bsp4j {
+    public static final String PROTOCOL_VERSION = "2.1.0";
```
lefou, line 4: This might be problematic, as this can be inlined by the Java compiler resulting an a potential wrong value at runtime. A static method can avoid this.

agluszak: Okay, I've just read Puzzle 93 from Bloch's book :) fixing it now

lefou: If I compile my library against version A, the compiler is allowed to inline the value "2.1.0" in the compiled binary code of the library. When I now use a newer version which would return "2.1.1", but do not recompile my library, the code which uses the value e.g. to construct a string, might still use the stale inlined "2.1.0" value. This is a general problem with constants in Java, hence constants which are suspected to change in a subsequent release (arbitrary but fixed), should not use a constant field but a constant method.

```suggestion
    public static final String protocolVersion() { return "2.1.0"; }
```

ckipp01 (approved):

jastice:
If there's no objections by Wednesday, let's merge this, okay?
@adpi2

build.sbt
```diff
@@ -56,6 +56,12 @@ import org.eclipse.xtend.core.XtendStandaloneSetup
 Global / cancelable := true
 publish / skip := true
 
+addCommandAlias(
+  "generate",
+  "codegen ; xtend ; scalafmtAll ; scalafmtSbt"
```
jastice, line 61:
```suggestion
  "; codegen ; xtend ; scalafmtAll ; scalafmtSbt"
```

jastice: sbt peculiarity, multi-commands need a `;` in front of the first

bsp4j/src/main/java/ch/epfl/scala/bsp4j/Bsp4j.java
Baccata: TIL 😄

PR #561, opened 2023-06-30
Make ScalaMainClass usable as both RunParamsData and DebugSessionParamsData

agluszak:
After the Smithy migration `ScalaMainClass` was no longer listed as a valid `RunParamsData`. Now the `dataKind` trait takes a list of "roles" a given structure can fulfill (instead of a single role).


jastice:
isn't this a breaking change to `dataKind` on protocol level, or am I misunderstanding something?

agluszak:
> isn't this a breaking change to dataKind on protocol level, or am I misunderstanding something?

@jastice No, it's not a breaking change at all, it restores things to the state before Smithy. It's about the `dataKind` trait in the smithy model, not the `dataKind` field in the protocol.

ckipp01 (approved):
LGMT!

PR #567, opened 2023-07-11
Less dependency bump PRs.

jastice:
Configure scala-steward and dependabot to do grouped dependency bump PRs on the first of each month. Generally this should be an appropriate interval as most updates are not time-critical. This change should decrease amount of notifications for maintainers somewhat.

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -25,7 +25,43 @@
 # Default: @asap
 #
 #pullRequests.frequency = "0 0 ? * 3" # every thursday on midnight
-pullRequests.frequency = "7 days"
+pullRequests.frequency = "0 0 0 1 * *" # At 12:00 AM, on day 1 of the month - to match dependabot `monthly` schedule
+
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):

.scala-steward.conf
```diff
@@ -25,7 +25,43 @@
...
+#
+# Every field in a `filter` is optional but at least one must be provided.
+#
+# For grouping every update together a filter like {group = "*"} can be # provided.
+#
+# To create a new PR for each unique combination of artifact-versions, include ${hash} in the name.
+#
+# Default: []
+pullRequests.grouping = [
+  { name = "major", "title" = "Major updates", "filter" = [ {"version" = "major"}] },
```
agluszak, line 62: I'm not sure if grouping major versions is a good idea

jastice: what's your objection? It's one of the defaults, I just took out the rest of the more granular ones.

PR #575, opened 2023-07-17
Split TaskData into TaskStartData, TaskProgressData and TaskFinishData

agluszak:
It makes more sense to keep these kinds separate


ckipp01 (approved):
Looks good to me LGTM

jastice:
I'm okay with merging it, but splitting them up like this kind of would suggest to me that these datakinds would go into different fields, even though they're meant to go in the same field.

jastice:
> What do you mean? They go into different fields in different structures

Ah, you're right. I remembered incorrectly. Doh!

PR #578, opened 2023-07-18
Restore the scala-attach-remote data kind

agluszak:
@ckipp01 @adpi2 could you please provide the description of this data kind? In the old doc [it wasn't very well documented](https://github.com/build-server-protocol/build-server-protocol/blob/v2.1.0-M4/docs/extensions/scala.md#scala-specific-data-kinds-for-debugging)
image

### DIFF
--- a/spec/src/main/resources/META-INF/smithy/bsp/extensions/scala.smithy
+++ b/spec/src/main/resources/META-INF/smithy/bsp/extensions/scala.smithy
@@ -258,7 +258,7 @@ list ScalaTestSuiteClasses {
     member: String
 }
 
-/// TODO
+/// The debug session will connect to a running process. The DAP client will send the port of the running process later.
 @dataKind(kind: "scala-attach-remote", extends: [DebugSessionParamsData])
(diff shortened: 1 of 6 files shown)


spec/src/main/resources/META-INF/smithy/bsp/extensions/scala.smithy
```diff
@@ -258,6 +258,11 @@ list ScalaTestSuiteClasses {
     member: String
 }
 
+/// TODO
```
adpi2, line 261:
```suggestion
/// An empty object. The `"scala-test-suites"` `dataKind` does not need any information.
```

agluszak: What I meant is to clarify what this data kind is used for. The specification says nothing about what should happen when a server receives a request with this data kind

adpi2: edited

PR #579, opened 2023-07-18
Simplify approval tests

agluszak:
Instead of an external library use the same mechanism we have for bsp4j and bsp4s.


ckipp01 (approved):
Good idea, this make sense. 👍🏼 Looks like there are some CI failures though around formatting, but I like the general idea.

adpi2 (approved):
LGTM

CONTRIBUTING.md
```diff
@@ -68,18 +68,9 @@ need to be replaced with the following 2 steps:
...
-In directory `codegen/src/test/scala` there will appear
-files `<testname>.<generated doc>.received.txt` for each generated doc which differs
-from the previously approved version. Approving means comparing `<file>.received.txt`
-with `<file>.approved.txt` and introducing changes to the `<file>.approved.txt` one.
-
-When you run `sbt test` again the tests should pass and `<file>.received.txt`
-will be removed.
+Upon committing changes to the protocol specification, a test is run to ensure that the generated
+protocol specification is identical to the committed one.
+If the test fails, please run `sbt generate` and commit the changes.
```
jastice, line 73: Isn't there some case where one would like to fix something?

agluszak: What do you mean?

jastice: I mean, in which cases should I do something other than just running generate and commit?

jastice: Never ever? Or just when nothing changes in the spec but the test breaks anyway?

agluszak: I'm not sure what you're referring to

agluszak: If nothing changes in the spec, but the test breaks anyway, that's a bug

PR #589, opened 2023-07-31
Fix incorrect closed string enum rendering in bsp4s

agluszak:
Fortunately we currently don't have closed string enums in the spec, so it doesn't change the generated code


ckipp01 (approved):
LGTM

PR #596, opened 2023-08-28
Add `Cancel` notification

KatKlo:
In JVM languages, cancelling requests is done by using the `CompletableFuture` feature, so the `Cancel` notification is unnecessary. However, some languages (such as Rust) don't support this feature. This means that servers/clients written in these technologies should have this notification in their bsp library, so we would like it to be explicitly defined in the protocol for proper generation process.


jastice (commented):
Thanks! I assume this is to add the cancel request explicitly to the protocol rather than implicitly via inheritance from LSP, but can you add a short explanation for the rationale in the PR description?

jastice (approved):

KatKlo:
I checked it by running intellij-bsp, and no `cancel notification` was sent. So no, no `cancel` under the hood for now. 

What's worse, since intellij-bsp doesn't send a `cancel notification`, there is no way to cancel a request while working with cargo-bsp. This `CompletableFuture` mechanism only works with servers that also use it, so it seems to work with bazel-bsp, but it won't work with servers like cargo-bsp.

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -640,6 +646,12 @@ structure TestProvider with [LanguageProvider] {
 
 }
 
+structure CancelRequestParams {
+    /// The request id to cancel.
+    @required
+    id: RequestId
```
agluszak, line 652: RequestId can be either a string or a number https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#cancelRequest

agluszak (changes requested):

ckipp01:
> > This `CompletableFuture` mechanism only works with servers that also use it, so it seems to work with bazel-bsp, but it won't work with servers like cargo-bsp.
> 
> That's not true. If you cancel a `CompletableFuture` a standard cancel notification (the same as described in this PR) is sent. https://github.com/eclipse-lsp4j/lsp4j/blob/main/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java#L149

This is what I went digging for as well and ended in the same place. Am I correct in assuming then that this isn't something we want to merge?

jastice:
> This is what I went digging for as well and ended in the same place. Am I correct in assuming then that this isn't something we want to merge?

imho we would still want to make it explicit so that BSP implementations don't need to rely on LSP libraries to supply all features implicitly

agluszak:
> imho we would still want to make it explicit so that BSP implementations don't need to rely on LSP libraries to supply all features implicitly

Exactly.

agluszak:
I think however that due to how cancellation is implemented in lsp4j this should be extracted to a separate interface (otherwise the default cancellation mechanism would be overridden). This can be done either in the lsp4j code generator or in the smithy definition (this would have to live in a separate "extension" then)

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -508,7 +508,11 @@ operation BuildTargetCleanCache {
 
 
 /// Represents the identifier of a BSP request.
-string RequestId
+@untaggedUnion
+union RequestId {
+    string: String
+    integer: Integer
+}
```
agluszak, line 515: instead of this breaking change I would suggest to rename `RequestId` to `OriginId` (as it is a concept introduced by BSP) and use `RequestId` only in `CancelParams`. This way `RequestId` would be the JsonRPC id

agluszak (approved):

spec/src/main/resources/META-INF/smithy/bsp/extensions/cancel.smithy
```diff
@@ -0,0 +1,33 @@
+$version: "2"
+
+namespace bsp.cancel
+
+use traits#jsonNotification
+use traits#jsonRPC
+use traits#untaggedUnion
+
+@jsonRPC
+service CancelExtension {
```
adpi2, line 10: Why is this an extension and not part of the core protocol?

KatKlo: @agluszak proposed this solution to prevent overriding lsp `Cancel` implementation and it seems reasonable

> I think however that due to how cancellation is implemented in lsp4j this should be extracted to a separate interface (otherwise the default cancellation mechanism would be overridden). This can be done either in the lsp4j code generator or in the smithy definition (this would have to live in a separate "extension" then)

adpi2: So this is a copy of the LSP cancel protocol.

Do we still depend on the lsp4j library? If so, do we have a plan to remove it?

agluszak: > Why is this an extension and not part of the core protocol?

Because some libraries don't have to implement it explicitly.

> So this is a copy of the LSP cancel protocol.

Yes

> Do we still depend on the lsp4j library?

Partially. The lsp4j library consists of two separate things. One is the JsonRpc library used by bsp4j. Cancellation handling is defined inside lsp4j.jsonrpc. The other thing is the support library for LSP (in the same sense as bsp4j is a Java support library for BSP). We never used it.

> If so, do we have a plan to remove it?

I don't think so.

adpi2: Ok so we need to implement cancellation in some languages but not in Java because we have lsp4j.

adpi2 (approved):

spec/src/main/resources/META-INF/smithy/bsp/extensions/cancel.smithy
agluszak: Exactly

PR #601, opened 2023-09-11
Fix "scala-test-suites" and "scala-test-suites-selection" data kinds

agluszak:
They were previously mistakenly marked as debug data kinds

### DIFF
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/DebugSessionParamsDataKind.java
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/DebugSessionParamsDataKind.java
@@ -3,6 +3,4 @@ package ch.epfl.scala.bsp4j;
 public class DebugSessionParamsDataKind {
     public static final String SCALA_ATTACH_REMOTE = "scala-attach-remote";
     public static final String SCALA_MAIN_CLASS = "scala-main-class";
-    public static final String SCALA_TEST_SUITES = "scala-test-suites";
-    public static final String SCALA_TEST_SUITES_SELECTION = "scala-test-suites-selection";
 }
--- a/bsp4j/src/main/java/ch/epfl/scala/bsp4j/TestParamsDataKind.java
+++ b/bsp4j/src/main/java/ch/epfl/scala/bsp4j/TestParamsDataKind.java
@@ -2,4 +2,6 @@ package ch.epfl.scala.bsp4j;
 
 public class TestParamsDataKind {
     public static final String SCALA_TEST = "scala-test";
+    public static final String SCALA_TEST_SUITES = "scala-test-suites";
+    public static final String SCALA_TEST_SUITES_SELECTION = "scala-test-suites-selection";
 }
--- a/bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
+++ b/bsp4s/src/main/scala/ch/epfl/scala/bsp/Bsp.scala
@@ -320,8 +320,6 @@ object DebugSessionParams {
 object DebugSessionParamsDataKind {
   val ScalaAttachRemote = "scala-attach-remote"
   val ScalaMainClass = "scala-main-class"
-  val ScalaTestSuites = "scala-test-suites"
-  val ScalaTestSuitesSelection = "scala-test-suites-selection"
 }
 
 final case class DependencyModule(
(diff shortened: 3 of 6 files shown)


jastice (approved):

bsp4j/src/main/java/ch/epfl/scala/bsp4j/DebugSessionParamsDataKind.java
```diff
@@ -3,6 +3,4 @@
 public class DebugSessionParamsDataKind {
     public static final String SCALA_ATTACH_REMOTE = "scala-attach-remote";
     public static final String SCALA_MAIN_CLASS = "scala-main-class";
-    public static final String SCALA_TEST_SUITES = "scala-test-suites";
```
ckipp01, line 6: Actually thinking back about this. I just checked and with Metals anyways if you trigger a debug tests, it will send this in, which will not currently break. How is it expected to debug test suites now?

CC @tgodzik

tgodzik: Hmm... Those are just names, no? So should basically be not much changes needed. 

Though I wonder why we have separate debug and test, since we could have everything in debug, but have a parameter that decides whether to debug when running or not.

ckipp01: Sure, but I was referring to places like this:

https://github.com/scalacenter/bloop/blob/abd12ecde91471b0caa5931ce2d720ed952d7bd3/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala#L627

Where we have debug session params that check for stuff like this, but no they are no longer there. Meaning that when you trigger a debug test in metals it will send debug session start with the params for tests... but that's no longer an option.

tgodzik: But we match on kind, which is a string, so it doesn't really matter where it is. Does it?

ckipp01: I mean of course we can just add a string... but that seems like a code smell to me that we're purposefully matching on the available values there when it's not a debug datakind. Meaning we're either not following convention, or we removed something we shouldn't. I'm not sure I'm explaining myself well here, so no worries.

tgodzik: > I mean of course we can just add a string... but that seems like a code smell to me that we're purposefully matching on the available values there when it's not a debug datakind.

We just moved the location of the string, so we just need to change import of the string we are matching on. This doesn't really change the data sent by the protocol, just the organization of the bsp library code. Whatever was sent before, will be sent still. The kind is just a string we add to the dataKind field. I am not sure I understand when this would break? This seems as a minor cosmetic change.

> So they are data kinds for debug, not for tests, after all? Or can they be used for both?

I think we always assumed that all kind can be used for debug. That is decided later on when initializing DAP. `TestParamsDataKind` might not be necessary as we don't have anything like that for Main classes

agluszak: > I think we always assumed that all kind can be used for debug. 

I don't know which "we" you mean, but no such thing was written in the specification, at least by the time I joined 😅
I mean, of course, all kinds can be used for anything, but the idea for writing them down in the spec was to inform how they are *supposed* to be used.

> TestParamsDataKind might not be necessary as we don't have anything like that for Main classes

I'm not sure if I understand

agluszak: I still don't understand. There is no `MainRequest`. Do you refer to [this one?](https://build-server-protocol.github.io/docs/extensions/jvm#buildtargetjvmrunenvironment-request) (which superseded `buildTarget/scalaMainClasses`)

tgodzik: >  There is no MainRequest.

That's what I mean, we have a special enum for tests and none for running main, we put that into debug kind. This is not really a problem for me, but if we are working on making it clearer, it would make sense to have separate main kind and test kind or just put everything into one.

Either way, I don't think it matters as those are just strings and if we migrate to a newer version of BSP I will just change the path to the particular kind.

tgodzik: ach right, I think I am getting lost with the different enums :sweat_smile:

PR #602, opened 2023-09-11
Improve running applications and tests using BSP

agluszak:
Expanded the functionality for running applications and tests using BSP. Added optional support for environment variables and working directory parameters. This allows the user to set up the specific environment for the tests or applications.

Arguments and environment variables lists in ScalaMainClass and ScalaTestSuites have been marked deprecated, because the base request now contains these arguments.


jastice:
From discussion:

- I have some concerns whether we want to move all process output through BSP, due to de/serialization overhead. We should verify that the overhead is acceptable.
- We may consider an alternative interface that just gives a reference to a process or file, even though that may increase client side complexity. 
- on the other hand, if we model a whole stdio through BSP as in this PR, that may increase complexity over passing the process reference to existing terminal utilities
- furthermore, we'd still not be handling non-stdio input/output (files or sockets)

> Lastly, arguments and environment variables lists in ScalaMainClass and ScalaTestSuites have been made optional and marked deprecated, because the base request now contains these arguments.

Making them optional would be a breaking change for clients that expect them to be defined.

agluszak:
> furthermore, we'd still not be handling non-stdio input/output (files or sockets)

I'm not sure what you mean by that. Files and sockets contents are not displayed in the terminal anyway so client has nothing to do with them I guess? Or do you mean redirecting the stdout to a file/socket or obtaining stdin from a file/socket? If the latter then the client will be able to do whatever it want with the contents once it has access to them

agluszak:
I've split off std streams handling to #604

agluszak:
> Making them optional would be a breaking change for clients that expect them to be defined.
reverted this one @jastice

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -1150,12 +1146,20 @@ structure CompileParams {
 
     /// Optional arguments to the compilation process.
     arguments: Arguments
+
+    /// Optional environment variables to set before compiling.
+    environmentVariables: EnvironmentVariables
```
adpi2, line 1151: (I don't think that can easily be supported by sbt or mill)

adpi2 (approved):
:+1:

spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
lefou: I wonder why environment variables are needed at all when compiling things? Compilers aren't typically configured via env vars. Most build tools don't have dedicated support for env vars that are passed to the compiler. And without such support, depending on env vars in the compilation would make it less reproducible. What is the motivation for that addition?

agluszak: > I wonder why environment variables are needed at all when compiling things? 

In most cases - they are not needed.

> Compilers aren't typically configured via env vars. 

Agree. But in some cases - they can be. For example: make can read env variables. In Bazel you can also [use env variables in the build process](https://bazel.build/reference/command-line-reference#flag--action_env). 

> What is the motivation for that addition?

Futureproofing. Ofc sbt & mill are free to ignore these arguments.

lefou: Yeah, in Mill you can also use env variables (and you can even track their value to capture any change), but I think that is not the point. The question is, do you need control over "different" env vars per compiler invocation? Where should these env vars come from? Do you expect special configuration for it in the BSP client? If it's a special configuration thing inside the build tools, shouldn't it be completely transparent to the client? 

We are not speaking about general stable env vars set in the environment of the user and staying the same while the BSP client is running, right?

agluszak: > Where should these env vars come from? Do you expect special configuration for it in the BSP client?

Yes, I imagine you could have some UI for it in a client

> We are not speaking about general stable env vars set in the environment of the user and staying the same while the BSP client is running, right?

Right. This would be used for overrides only.

agluszak: But in general I'm not particularly attached to this specific addition, as I said I added it for future-proofing and consistency with other requests. But we have no plans to implement compile-specific env var settings in bazel-bsp. So I'm 100% fine with cutting this out and revisiting it if anyone needs it later.

lefou: My guess, this is the kind of future-proofing that produces more work and complexity, but never pays out. But you it appears to me you have a real use-case in mind, so I'll mute. The spec should be clear whether this feature is optional or not for BSP servers, as it makes compilation tasks quasi uncachable.

agluszak: I'll just remove this part and if we decide to implement it I'll come back :)

PR #604, opened 2023-09-13
Add functionality for reading stdin and printing stdout/stderr

agluszak:
This commit introduces new methods in BuildServer and BuildClient classes that allow for interaction with the current running target. A 'PrintParams' class is also added to encapsulate parameters for these methods. The added methods are 'onRunReadStdin' in BuildServer, and 'onRunPrintStdout' and 'onRunPrintStderr' in BuildClient. These changes allow passing user input to the target's stdin stream and receiving its stdout and stderr streams. The documentation and specifications have been updated accordingly.


ckipp01:
So this might be an adjacent question to this change, but reading through the notes on this I kept thinking about DAP. I think we should be careful to not venture into the territory of duplicating parts of things that are included and meant to be in the DAP world instead of BSP. Where is the line between using DAP to get the output from running vs this change? I don't fully understand the usecase for this.

agluszak:
DAP is about debugging, while this PR has nothing to do with debugging. We could of course inherit the [output event from DAP](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Output), but then it would be used **outside of the DAP session**. 

> I don't fully understand the usecase for this.

When you run a runnable target (or test a testable target) the application being run can produce output. We would like to be able to display this output in a run/test console in IntelliJ. Currently there's no good endpoint for sending this output from the server to the client. We use `build/logMessage`, but in our opinion this endpoint should only be used for logging, well, logs. Logs of the server. We could use task progress messages, but again in our opinion this is not the best tool for the job. There's also no way to pipe user input in the other direction: when the user types something in the console we would like to be able to send this input through BSP to the program being run.

ckipp01:
> DAP is about debugging, while this PR has nothing to do with debugging. We could of course inherit the [output event from DAP](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Output), but then it would be used **outside of the DAP session**.
> 
> > I don't fully understand the usecase for this.
> 
> When you run a runnable target (or test a testable target) the application being run can produce output. We would like to be able to display this output in a run/test console in IntelliJ. Currently there's no good endpoint for sending this output from the server to the client. We use `build/logMessage`, but in our opinion this endpoint should only be used for logging, well, logs. Logs of the server. We could use task progress messages, but again in our opinion this is not the best tool for the job. There's also no way to pipe user input in the other direction: when the user types something in the console we would like to be able to send this input through BSP to the program being run.

I think part of my confusion comes from our usecase in Metals. For example even when we just do a run we start a debug session or we run it as a process I believe. I'm actually not positive with the recent changes how it works under the hood. Pinging @tgodzik since this might have an impact on how things are done in Metals related to this.

tgodzik:
Currently there are two ways of running in Metals, either via DAP and starting a session or running the java command directly in terminal.

> Currently there's no good endpoint for sending this output from the server to the client. We use build/logMessage, but in our opinion this endpoint should only be used for logging, well, logs. Logs of the server. 

 We actually want to try and also run within server and forward the output. I am using logMessage for that, but it's not great since some messages might get mixed in. So this change might be useful. Potentially we could make it a bit simpler with a separate event and an originId to tie it to the run event?

tgodzik:
> > Potentially we could make it a bit simpler with a separate event and an originId to tie it to the run event?
> 
> You mean introducing a new kind of `MessageType`? https://build-server-protocol.github.io/docs/specification/#onbuildshowmessage-notification

Maybe? Plus might make sense to use `build/showMessage` for it instead of logMessage. This wouldn't require almost any changes in the build servers.

We could have:
```
/** A standard output message. */
  Output = 4,
/** A standard error message. */
  ErrorOutput = 5,
```

agluszak:
> This wouldn't require almost any changes in the build servers.

I don't think the change would be that big anyway. Using `build/showMessage` won't help with stdin, too.

```
export interface ShowMessageParams {
  type: MessageType;

  task?: TaskId;

  originId?: RequestId;

  message: string;
}
```

```
export interface PrintParams {
  originId: Identifier;

  task?: TaskId;

  message: string;
}
```
```

File: bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java [hunk 1]
```diff
@@ -90,5 +90,19 @@ public interface BuildClient {
     @JsonNotification("build/taskFinish")
     void onBuildTaskFinish(TaskFinishParams params);
 
+    /**
+     * Notification sent from the server to the client when the target being run prints
```
adpi2, line 94: Should it also be used during test?

```suggestion
     * Notification sent from the server to the client when the target being run or tested prints
```

bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildServer.java
```diff
@@ -189,5 +189,12 @@ public interface BuildServer {
     @JsonRequest("buildTarget/cleanCache")
     CompletableFuture<CleanCacheResult> buildTargetCleanCache(CleanCacheParams params);
 
+    /**
+     * Notification sent from the client to the server when the user wants to send
+     * input to the stdin of the running target.
+     */
+    @JsonNotification("run/readStdin")
+    void onRunReadStdin(PrintParams params);
```
adpi2, line 197: For naming consistency, I think it would be better to use `ReadParams` instead of `PrintParams`, at the cost of duplication.

README.md
```diff
@@ -37,3 +37,9 @@ _Past Maintainers_
 
 - Jorge Vicente Cantero - [@jvican](https://github.com/jvican)
 - Ólafur Páll Geirsson - [@olafurpg](https://github.com/olafurpg)
+
+## Protocol Changelog
+
+### Unreleased
+
+- Add unstable `run/printStdout`, `run/printStderr` and `run/readStdin` notifications
```
adpi2, line 45: What do you mean by unstable? What is the process to make them stable?

adpi2 (commented):
As raised by @ckipp01, there is some overlap between this and DAP, but I can understand that we don't want to depend on all DAP for running and testing. As long as we keep the test and run requests, the protocol should handle stdin and stdout properly.

README.md
jastice: We introduced `unstable` additions to the protocol for new changes that haven't been tested yet on implementations. The expectation is that they may change in incompatible ways even in minor versions. We mark them stable when there is consensus that they work as they are in real world implementations

agluszak: Unstable means that they are exempt from semver stability guarantee. The process to make them stable is to wait some time, see if we need to introduce some changes and remove the @unstable annotation

bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildClient.java
```diff
@@ -90,5 +90,19 @@ public interface BuildClient {
...
 
+    /**
+     * Notification sent from the server to the client when the target being run or tested prints
+     * something to stdout.
+     */
+    @JsonNotification("run/printStdout")
+    void onRunPrintStdout(PrintParams params);
+
+    /**
+     * Notification sent from the server to the client when the target being run prints
```
adpi2, line 101:
```suggestion
     * Notification sent from the server to the client when the target being run or tested prints
```

agluszak:
@adpi2 I introduced the changes you suggested

adpi2 (approved):

PR #611, opened 2023-09-20
Fix `deprecated` trait usages

KatKlo:


codegen/src/main/scala/bsp/codegen/docs/TypescriptRenderer.scala
```diff
@@ -150,18 +150,23 @@ class TypescriptRenderer {
...
-      Lines(lines.toList)
+      lines(lastIndex) = lines(lastIndex) + " */"
+    }
+    Lines(lines.toList)
+  }
+
+  def renderDocumentation(hints: List[Hint]): Lines = hints
+    .collectFold {
+      case Documentation(string) => renderComments(string)
+      case Deprecated(message)   => renderComments(s"@deprecated $message")
```
agluszak, line 169: That's incorrect, as now if there's both a Documentation hint and a Deprecated hint two comment blocks will be rendered

agluszak (changes requested):

codegen/src/main/scala/bsp/codegen/docs/TypescriptRenderer.scala
```diff
@@ -150,18 +150,21 @@ class TypescriptRenderer {
...
-        val lastIndex = lines.length - 1
-        for (i <- 1 to lastIndex) {
-          lines(i) = " * " + lines(i)
-        }
-        lines(lastIndex) = lines(lastIndex) + " */"
+  def renderDocumentation(hints: List[Hint]): Lines = {
+    val lines = hints.flatMap {
+      case Documentation(string) => string.split(System.lineSeparator())
+      case Deprecated(message)   => Array(s"@deprecated $message")
+      case _                     => Array.empty[String]
```
agluszak, line 157:
```suggestion
      case Deprecated(message)   => Array(s"Deprecated: $message")
      case Unstable => Array("Unstable: may change in the future")
```

agluszak (changes requested):

agluszak (approved):

agluszak:
We have to merge https://github.com/build-server-protocol/build-server-protocol/pull/612 first

PR #612, opened 2023-09-22
Bump protocol version

agluszak:
Should also fix the failing CI


website/generated/docs/bindings/java.md
```diff
@@ -16,7 +16,7 @@ Add the following snippet to your build to add dependency on `bsp4j`.
 ### Gradle
 
 ```groovy
-compile group: 'ch.epfl.scala', name: 'bsp4j', version: '2.1.0-M6'
+compile group: 'ch.epfl.scala', name: 'bsp4j', version: '2.2.0-M1.alpha'
```
adpi2, line 19: Where does this `-M1.alpha` comes from?

website/generated/docs/extensions/cargo.md
```diff
@@ -9,7 +9,7 @@ protocol.
 
 ## BSP version
 
-`2.1.0`
+`2.2.0`
```
adpi2, line 12: When will this change be published to the website?

adpi2 (commented):
A few questions to compensate for my ignorance

agluszak:
@adpi2 I made a pre-release a few days ago with the newest changes: https://github.com/build-server-protocol/build-server-protocol/releases/tag/v2.2.0-M1.alpha

website/generated/docs/bindings/java.md
agluszak: from the newest tag on git

agluszak:
@adpi2 can you accept the PR? or do you have further questions?

jastice:
I don't think we need `.alpha` on top of `M1`, which is already meant as intermediate / pre-release

agluszak:
M1 is meant as intermediate/pre-release? It looks like so far we've been using versioning scheme `{protocol_version}-M{library_version}`

jastice:
> M1 is meant as intermediate/pre-release? It looks like so far we've been using versioning scheme {protocol_version}-M{library_version}

No, `M-x` was always just an incremented number before a "proper" release.

lefou:
`2.1.0-M5` means "Milestone 5" toward a `2.1.0` release.

agluszak:
Okay, so maybe let's discuss the versioning scheme we want to use from now. It looks like we have never had a proper 2.1.0 release then. The new additions to the protocol should land in 2.2.0 IMO. And we had 2.1.0-M6 release which didn't contain any breaking changes to the protocol, but broke some stuff in the libraries because of being auto-generated.

lefou:
We could tag `2.1.0-M5` also as `2.1.0` making the version which is currently used/implemented by most tools a stable one. 

After that, we could start to split the versions.

Next BSP version is `2.2.0`, if it's not breaking with `2.1.0`, otherwise `3.0.0`.

For libraries, we take the classic semantic versioning approach, based on the ABI. Each library has a programmatic way to provide the implemented BSP version. Question is, do we need to encode the implemented version in the library name? I think not, as long as the chanelog states which versions match. Or we provide a version matrix in the documentation.

jastice (approved):

PR #614, opened 2023-09-25
Update `Diagnostic` related structures

KatKlo:


spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy
```diff
@@ -712,37 +706,44 @@ structure Diagnostic {
...
 }
 
 structure Position {
-    /// Line position within a file. First line of a file is 0.
+    /// Line position in a document (zero-based).
     @required
     line: Integer
-    /// Character position within a line. First character of a line is 0.
+    /// Character offset on a line in a document (zero-based). The meaning of this
+    /// offset is determined by the negotiated `PositionEncodingKind`.
```
agluszak, line 734: > PositionEncodingKind

we don't have that in BSP

agluszak (approved):

PR #633, opened 2023-12-06
`javaVersion` doc in `JvmBuildTarget` includes javac `-target` flag

abrams27:
hi,

one small (but important) change in the docs for `JvmBuildTarget` - imo it's the way how it should be used, or maybe I'm missing something here?


agluszak:
@tgodzik @adpi2 @lefou any thoughts on this?

tgodzik:
Sounds reasonable, we don't actually use that field other than for displaying it currently.

agluszak (approved):

PR #650, opened 2024-02-05
improvement: Move classpath to a separate request

tgodzik:
More context https://github.com/JetBrains/bazel-bsp/pull/519#issuecomment-1926710910

### DIFF
--- a/spec/src/main/resources/META-INF/smithy/bsp/extensions/java.smithy
+++ b/spec/src/main/resources/META-INF/smithy/bsp/extensions/java.smithy
@@ -50,7 +50,7 @@ structure JavacOptionsItem {
     /// the -classpath flag in the command line interface
     /// of javac.
     @required
-    @deprecated(message: "This might be empty, use the buildTarget/jvmCompileClasspath endpoint")
+    @deprecated(message: "When the jvmCompileClasspath capability is supported, this might be empty and the buildTarget/jvmCompileClasspath endpoint should be used instead.")
     classpath: Classpath
     /// The output directory for classfiles produced by this target
     @required
--- a/spec/src/main/resources/META-INF/smithy/bsp/extensions/java.smithy
+++ b/spec/src/main/resources/META-INF/smithy/bsp/extensions/java.smithy
@@ -50,6 +50,7 @@ structure JavacOptionsItem {
     /// the -classpath flag in the command line interface
     /// of javac.
     @required
+    @deprecated(message: "This might be empty, use the buildTarget/jvmCompileClasspath endpoint")
     classpath: Classpath
     /// The output directory for classfiles produced by this target
     @required
--- a/spec/src/main/resources/META-INF/smithy/bsp/extensions/jvm.smithy
+++ b/spec/src/main/resources/META-INF/smithy/bsp/extensions/jvm.smithy
@@ -115,4 +115,38 @@ list JvmOptions {
 }
 
 
+/// The build target classpath request is sent from the client to the server to
+/// query the target for its compile classpath.
+@jsonRequest("buildTarget/jvmCompileClasspath")
+operation BuildTargetClasspath {
+    input: ClasspathParams
+    output: ClasspathResult
+}
+
+
+structure ClasspathParams {
+    @required
+    targets: BuildTargetIdentifiers
+}
+
+structure ClasspathResult {
+    @required
+    items: ClasspathItems
+}
+
+structure ClasspathItem {
+    @required
+    target: BuildTargetIdentifier
+    /// The dependency classpath for this target, must be
+    /// identical to what is passed as arguments to
+    /// the -classpath flag in the command line interface
+    /// of scalac.
+    @required
+    classpath: Classpath
+}
+
+list ClasspathItems {
+    member: ClasspathItem
+}
+
 
--- a/spec/src/main/resources/META-INF/smithy/bsp/extensions/jvm.smithy
+++ b/spec/src/main/resources/META-INF/smithy/bsp/extensions/jvm.smithy
@@ -118,23 +118,23 @@ list JvmOptions {
 /// The build target classpath request is sent from the client to the server to
 /// query the target for its compile classpath.
 @jsonRequest("buildTarget/jvmCompileClasspath")
-operation BuildTargetClasspath {
-    input: ClasspathParams
-    output: ClasspathResult
+operation BuildTargetJvmCompileClasspath {
+    input: JvmCompileClasspathParams
+    output: JvmCompileClasspathResult
 }
 
 
-structure ClasspathParams {
+structure JvmCompileClasspathParams {
     @required
     targets: BuildTargetIdentifiers
 }
 
-structure ClasspathResult {
+structure JvmCompileClasspathResult {
     @required
-    items: ClasspathItems
+    items: JvmCompileClasspathItems
 }
 
-structure ClasspathItem {
+structure JvmCompileClasspathItem {
     @required
     target: BuildTargetIdentifier
     /// The dependency classpath for this target, must be
@@ -145,8 +145,8 @@ structure ClasspathItem {
     classpath: Classpath
 }
 
-list ClasspathItems {
-    member: ClasspathItem
+list JvmCompileClasspathItems {
+    member: JvmCompileClasspathItem
 }
 
 
--- a/spec/src/main/resources/META-INF/smithy/bsp/extensions/scala.smithy
+++ b/spec/src/main/resources/META-INF/smithy/bsp/extensions/scala.smithy
@@ -103,7 +103,7 @@ structure ScalacOptionsItem {
     /// the -classpath flag in the command line interface
     /// of scalac.
     @required
-    @deprecated(message: "This might be empty, use the buildTarget/jvmCompileClasspath endpoint")
+    @deprecated(message: "When the jvmCompileClasspath capability is supported, this might be empty and the buildTarget/jvmCompileClasspath endpoint should be used instead.")
     classpath: Classpath
     /// The output directory for classfiles produced by this target
     @required
--- a/spec/src/main/resources/META-INF/smithy/bsp/extensions/scala.smithy
+++ b/spec/src/main/resources/META-INF/smithy/bsp/extensions/scala.smithy
@@ -103,6 +103,7 @@ structure ScalacOptionsItem {
     /// the -classpath flag in the command line interface
     /// of scalac.
     @required
+    @deprecated(message: "This might be empty, use the buildTarget/jvmCompileClasspath endpoint")
     classpath: Classpath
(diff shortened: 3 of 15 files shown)


lefou:
My first reaction was, that it's nice that we can defer the classpath request until it's really needed. 

Yet, I wonder what the potential impact might be, when we move the scalac options up into the buildTargetData, as you currently proposed by your change? Technically, they are not as important, that they need to be  provided that early. 
Compilation is done by the build server, so the client only needs the options for other probably optional or advanced stuff. Calculating scalac option may include some potential heavy work, e.g. when there are compiler plugins involved. With your change, the build client will trigger this with the buildTargets request, which is very early and can't be omitted.

The other thing is that the javac options now seem to end in the JvmBuildData, which is technically independent of the Java extension. This doesn't make sense to me.

I propose the keep the dedicated javac- and scalac options requests. The extraction of the classpath by itself should already improve the responsiveness.

adpi2:
> However, I don't think there is a needed to only send scalacOptions in a completely separate request. I think it actually adds complexity.

> I propose the keep the dedicated javac- and scalac options requests. The extraction of the classpath by itself should already improve the responsiveness.

I kind of agree here. To get the scalac options we need to resolve the compiler plugins which can be slow or fail (network issues). At the same time, `ScalaBuildTargetData` contains the `jars` of the scala-library and compiler, which already triggers the resolution. So adding the scalac options to `ScalaBuildTargetData` does not change anything in practice. (In sbt there is one global resolution for each project, that resolves the compiler, the scala-library, the compiler plugins, and the libraries)

Ideally I think we should  remove the `jars` from `ScalaBuildTargetData` and keep the `scalacOptions` out of it: the `ScalaBuildTargetData` should not trigger any resolution.

tgodzik:
> I propose the keep the dedicated javac- and scalac options requests. The extraction of the classpath by itself should already improve the responsiveness.

Sure, that was the other option I considered, but this seemed safer in terms of compatibility and migration. Removing classpath from scalacOptions would break binary compatibility, so I wasn't sure how to go around it.

Any ideas around there

> Ideally I think we should remove the jars from ScalaBuildTargetData and keep the scalacOptions out of it: the ScalaBuildTargetData should not trigger any resolution.

Metals doesn't use jars neither does scala-cli, so this should be safe to remove,  but not sure where it should end up?

Any ideas about not breaking compatibility here? Or just deprecate it to be removed in 3.x ?

tgodzik:
We could also have `resolveClasspathRequest` capability for both client and server so that they would know to send the empty deprecated fields. :thinking:

adpi2:
> Metals doesn't use jars neither does scala-cli, so this should be safe to remove, but not sure where it should end up?

I am pretty sure that IntelliJ needs it. Can you confirm @jastice?

> Any ideas about not breaking compatibility here? Or just deprecate it to be removed in 3.x ?

Probably we need to deprecate it to be removed in 3.x.

lefou:
From a compatibility perspective, the classpath field can just be kept empty (not ideal but better that waiting for a new major version, IMHO). The fact that the classpath request should be used (instead) can be either derived from the implemented BSP version or communicated via a capability. In other words, if you're going to implement and support the new classpath request, just leave the deprecated classpath field empty. The bsp4j impl could even pre-initialize it.

tgodzik:
Ok, I changed to just add the new method. Should we add a capability?

adpi2:
> Ok, I changed to just add the new method. Should we add a capability?

I would say yes.

tgodzik:
> > Ok, I changed to just add the new method. Should we add a capability?
> 
> I would say yes.

Done!

lefou:
Can you a bit more specific in what exact cases the `classpath` is going to be used? Is it four compilation (only)? Is it also supposed to be used at runtime? It should, as it will most probably break the idea of a run classpath most build tools implement in some way?

tgodzik:
> Can you be a bit more specific in what exact cases the `classpath` is going to be used? Is it for compilation (only)? Is it also supposed to be used at runtime? It shouldn't, as it will most probably break the idea of a run classpath most build tools implement in some way.

We currently mostly use classpath for the presentation compiler, so that would be compile classpath. Should we have the two here? Although, we get runtime classpath in the jvmEnvironment already

lefou:
> > Can you be a bit more specific in what exact cases the `classpath` is going to be used? Is it for compilation (only)? Is it also supposed to be used at runtime? It shouldn't, as it will most probably break the idea of a run classpath most build tools implement in some way.
> 
> We currently mostly use classpath for the presentation compiler, so that would be compile classpath. Should we have the two here? Although, we get runtime classpath in the jvmEnvironment already

Thanks for the clarification. I don't think we need a extra run classpath here. But I'm a bit unsure if we should rename the `classpath` to compileClasspath` for clarity.

tgodzik:
> > > Can you be a bit more specific in what exact cases the `classpath` is going to be used? Is it for compilation (only)? Is it also supposed to be used at runtime? It shouldn't, as it will most probably break the idea of a run classpath most build tools implement in some way.
> > 
> > 
> > We currently mostly use classpath for the presentation compiler, so that would be compile classpath. Should we have the two here? Although, we get runtime classpath in the jvmEnvironment already
> 
> Thanks for the clarification. I don't think we need a extra run classpath here. But I'm a bit unsure if we should rename the `classpath` to `compileClasspath` for clarity.

That makes sense, @adpi2 what do you think?

adpi2:
> Thanks for the clarification. I don't think we need a extra run classpath here. But I'm a bit unsure if we should rename the classpath to compileClasspath for clarity.

I agree. Also I think we should rename them `jvmCompileClasspath` and `jvmCompileClasspathProvider` to make it clear they belong to the JVM extension.

jastice:
>> Metals doesn't use jars neither does scala-cli, so this should be safe to remove, but not sure where it should end up?
> I am pretty sure that IntelliJ needs it. Can you confirm @jastice?

Yes, IntelliJ Scala BSP implementation does use `jars`

tgodzik:
> > Thanks for the clarification. I don't think we need a extra run classpath here. But I'm a bit unsure if we should rename the classpath to compileClasspath for clarity.
> 
> I agree. Also I think we should rename them `jvmCompileClasspath` and `jvmCompileClasspathProvider` to make it clear they belong to the JVM extension.

Done!

> > > Metals doesn't use jars neither does scala-cli, so this should be safe to remove, but not sure where it should end up?
> > > I am pretty sure that IntelliJ needs it. Can you confirm @jastice?
> 
> Yes, IntelliJ Scala BSP implementation does use `jars`

I will skip it for this PR then.

spec/src/main/resources/META-INF/smithy/bsp/extensions/java.smithy
```diff
@@ -50,6 +50,7 @@ structure JavacOptionsItem {
     /// the -classpath flag in the command line interface
     /// of javac.
     @required
+    @deprecated(message: "This might be empty, use the buildTarget/jvmCompileClasspath endpoint")
```
lefou, line 53:
```suggestion
    @deprecated(message: "When the jvmCompileClasspath capability is supported, this might be empty and the buildTarget/jvmCompileClasspath endpoint should be used instead.")
```

Or anything else, that makes clear, when this can be empty.

spec/src/main/resources/META-INF/smithy/bsp/extensions/scala.smithy
```diff
@@ -103,6 +103,7 @@ structure ScalacOptionsItem {
     /// the -classpath flag in the command line interface
     /// of scalac.
     @required
+    @deprecated(message: "This might be empty, use the buildTarget/jvmCompileClasspath endpoint")
```
lefou, line 106:
```suggestion
    @deprecated(message: "When the jvmCompileClasspath capability is supported, this might be empty and the buildTarget/jvmCompileClasspath endpoint should be used instead.")
```

spec/src/main/resources/META-INF/smithy/bsp/extensions/jvm.smithy
```diff
@@ -115,4 +115,38 @@ list JvmOptions {
 }
 
 
+/// The build target classpath request is sent from the client to the server to
+/// query the target for its compile classpath.
+@jsonRequest("buildTarget/jvmCompileClasspath")
+operation BuildTargetClasspath {
```
adpi2, line 121: For consistency, I suggest to rename these data types to `BuildTargetJvmCompileClasspath`, `JvmCompileClasspathParams` etc.

spec/src/main/resources/META-INF/smithy/bsp/extensions/scala.smithy
tgodzik: Done, thanks!

spec/src/main/resources/META-INF/smithy/bsp/extensions/jvm.smithy
tgodzik: Makes sense, renamed

tgodzik:
Sorry for the delay, I was off for the last week. It should now be fixed, should we also wait for @jastice approve?

adpi2 (approved):

lefou (approved):

tgodzik:
@jastice what do you think about merging it? Would be also useful to have an additional 2.2.x with this before the big changes hit. This would make it much easier to handle Bazel and Mill.

lefou:
> This would make it much easier to handle Bazel and Mill.

I think it would also improve sbt-bsp, but we might just not have witnessed these more complex builds there.

agluszak:
> @jastice what do you think about merging it? Would be also useful to have an additional 2.2.x with this before the big changes hit. This would make it much easier to handle Bazel and Mill.

let's merge it!

lefou:
Or could this somehow be handled in the bsp4j lib?

lefou:
To this moment, we never had changes, which removed some necessary work from a request. So the client version was more or less ignored on the Mill server side.

tgodzik:
We would need client capability for that too to make sure classpath is being used :thinking: 

I think the plan would be:
1. Release 2.2.0-M2 with these changes
2. Test it out with Bazel BSP + Metals
3. If any issues arise fix them here
4. Add the client capability in the meantime

Anyone against the plan?

lefou:
If we know, we need the client capability to properly implement it, why not add it before the next release? Otherwise, the servers need to _assume_ that the client does or does not support it, which isn't ideal. But more importantly, such a prepared test-version of a server will not properly work for all regular clients. So, you could do these test only with dedicated servers and clients.

tgodzik:
> If we know, we need the client capability to properly implement it, why not add it before the next release?

Sure, that's probably the saner approach :sweat_smile: I revised the plan.

tgodzik:
Raised here https://github.com/build-server-protocol/build-server-protocol/pull/656

PR #680, opened 2024-06-14
More prominently advertise contributing entities

jastice:


adpi2 (approved):
It looks great! Thanks.

lefou:
So, "contributing" means pull request?

jastice:
> So, "contributing" means pull request?

I haven't thought about the precise definition, but that would certainly be included

PR #690, opened 2024-08-20
Add build tool params to compile/run/test

agluszak:


lefou:
Is there some preceding discussion about this change, somewhere?

agluszak:
> Is there some preceding discussion about this change, somewhere?

No, let's discuss here

adpi2 (commented):
It seems this pattern is common enough to be added to BSP.

However this won't be supported in sbt, because the BSP server process and the sbt process are the same, and it would not make sense for it to restart itself each time the `buildToolArguments` change.

So I suggest to add some sort of capability in `BuildServerCapabilities` to denote whether the server supports the feature or not.

agluszak:
> Also, these protocol changes look more like user specific or use-case specific arguments, not build-tool specific arguments. 

So do you disagree with the naming here or what's your point?

> If there is something specific to a build tool, it should be configured when initializing the build tool to stay stable afterwards for all build targets. Applying them to each single run/test/compile invocation looks like unnecessary overhead and a source of inconsistencies.

But why? You might want to have a button in the IDE which, for example, builds all cargo targets with `--keep_going` and one which doesn't. It's a per-compile-request setting.

> since compile is a prerequisite of the other and should be in best case reproducible. Depending on some temporary arguments is the opposite of reproducible.

If the user wants to provide arguments which make the build non-reproducible it's their problem, not ours. But there are many use cases which don't affect the reproducibility at all. See the `--keep_going` example from Cargo (Bazel support exactly the same scenario) or [using bazel to prepare a release](https://bazel.build/docs/user-manual#significant-options) or [setting an execution strategy](https://bazel.build/reference/command-line-reference#flag--strategy) or [toggling the configuration cache/parallel building in Gradle](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_performance)

lefou:
I'm still trying to understand the use case. But I think the `--keep-going` example is a good one. It doesn't affect the quality of the outcome, but the `quantity`. So the idea is to give the build tool some hints, which do not affect the result items individually. (It could be, this is a symptom of the specific nature of BSP to bundle up bulk requests rather then address individual build targets.)

I think what's missing is to add some motivation/example in the spec, to the user better known for what kind of options this feature is.

I also agree with @adpi2 that a `BuildServerCapability` makes sence. We should specify explicitly how a build server should report unsupported build tools params.

PR #698, opened 2024-09-05
[maintenance] Bump deps

agluszak:


MODULE.bazel.lock
adpi2, line 1: Do you know why this file is so much smaller now?

adpi2: It has lost 11,000 lines which seems suspicious to me

agluszak: Thanks for vigilance! I found it suspicious at first too, but I think it's because of https://github.com/bazelbuild/rules_jvm_external/pull/1141, https://github.com/bazelbuild/bazel/pull/22351 and maybe some other change on the Bazel side

adpi2 (approved):

PR #708, opened 2024-12-09
fix the branch with bumps for release

hb-man:
helps fix branch with bumps
made some more updates to existing bumps in the branch
restored removed lines in Scala files
bumped more dependencies and bazel version


hb-man:
@agluszak please take a look

